PR #50, opened 2014-10-22
LibraryVariableSet Import and Export

kpitt:
When importing and exporting a project, the project-level variables are included. However, if the project references a library variable set, there is no command to import and export the variables and values in the shared set. This changes adds an importer and exporter of type "libraryvariableset" to handle these shared variable values.

The underlying VariableSet that stores the values is the same for LibraryVariableSet as for Project, so most of the code for import and export is copied from ProjectImporter. As a result, there is a lot of duplication in the code, especially in the more complicated import logic. However, I wanted to keep it simple for this initial request and avoid modifying any existing code.

I have an additional [variable-import-refactor](https://github.com/kpitt/Octopus-Tools/commits/variable-import-refactor) branch that builds on this change by refactoring ProjectImporter and LibraryVariableSetImporter to share the common code.


kpitt:
After working with this for awhile, I realized that overwriting the entire library variable set with the set of variables in the import file can cause real problems when exporting from one server and importing into another. To make it easier to share variable updates, I implemented a merge process that will preserve  existing variables that have a different scope than the variables in the import file. This allows new variables to be imported into the library without destroying values that have been scoped to unique environments on the target server.

Note that we still need to reset the values of existing sensitive variables to an empty string when they are preserved. The Octopus API requires that we build a complete variable collection on the client side to send back in the update request. We can't retrieve the value of a sensitive variable for security reasons, so it is impossible to preserve that value in the update request.

PaulStovell:
Again, thanks for this PR, and sorry we failed at reviewing it in time. The new migrator can export from two different servers as JSON (and supports sensitive values), so merging them should be somewhat automatable.

PR #54, opened 2015-02-03
Default lifecycle on import

alfhenrik:


DamianMac:
We should export the lifecycle id and name as part of the export. Then when we import try to match the lifecycle name in the target machine, failing that we should go to the default lifecycle. As it stands this would work for a 2.5 -> 2.6 import, but would fail on a 2.6 -> 2.6 import.

PR #55, opened 2015-02-27
Fix typo in import command options group name

kpitt:
Minor cosmetic change to fix the label used for the command-specific options on the "import" command.


alfhenrik:
Ah, the good ol' ctrl-c, ctrl-v typo :blush: Thanks for finding and fixing :+1:

PR #56, opened 2015-03-16
Warn about sensitive step properties on import

kpitt:
We already have warnings when importing sensitive variables so the
user knows that the actual values were not imported.  We need
similar warnings for deployment process steps because it's easy to
miss sensitive settings such as a custom service user password.

Unlike variables, the UI only shows a descriptive label for the
step properties, so the actual property name would not be useful
in the log message.  There are also properties that may not be
used depending on the values of other related properties, so they
don't really need to be set.  Instead, just log a general warning
for each step that contains sensitive properties so that the
user knows to check those steps after the import is complete.


PaulStovell:
Thanks for the pull request and sorry it took so long to take a look at it. We ended up building a fully-fledged migration tool for 3.0, and these import/export tools are going to be obsolete. For that reason I'll close this PR.

PR #57, opened 2015-05-11
Added DryRun feature and refactored importers

Damovisa:
Also reports on issues with import a bit more kindly. All issues will get reported rather than crashing on the first.


source/OctopusTools/Importers/CheckedReferences.cs
```diff
@@ -0,0 +1,43 @@
...
+        public ICollection<string> MissingDependencyNames { get; set; }
+
+        public List<string> FoundDependencyIds
+        {
+            get { return FoundDependencies.Keys.ToList(); }
+        }
+
+        public IEnumerable<string> MissingDependencyErrors
+        {
+            get { return MissingDependencyNames.Select(d => string.Format("Could not find {0} '{1}'", typeof (T).Name.Replace("Resource",""), d)); }
```
zentron: If this is generic type <T> where T is always the same, can we resolve its name once rather than in select iteration?

zentron:
looks good :+1:

PR #58, opened 2015-05-12
When searching for prerelease package, take out the partialmatch opti…

vermeeca:
…on and bring back multiple versions of a specific package

Fixes Issue [1595](https://github.com/OctopusDeploy/Issues/issues/1595)


PaulStovell:
Like #59, I think this will be solved by branching support.

PR #59, opened 2015-06-12
Prerelease fallback

vermeeca:
This PR builds upon [PR 58](https://github.com/OctopusDeploy/Octopus-Tools/pull/58), so I'm not sure of the proper protocol for PRs like this.  If I need to re-submit for some reason, just LMK.

Anyway, this PR adds an extra parameter to the CreateRelease command that lets the command fall-back to the latest version of a package, if no pre-release version exists.

I found that I wanted this feature if I had a project that deploys multiple packages and I wanted to create a release for which only one of those packages was a pre-release version.


PaulStovell:
Thanks for the PR, but for now I'm going to close it. Octo.exe is getting too many arguments for specific cases. Our upcoming feature for branching support, which will let you specify semver ranges, should fix these kinds of issues in a more generic (not just Octo.exe) way.

PR #60, opened 2015-08-20
Octo.exe displays username that owns API key and suggests permissions issues

IanMercer:
Fixed OctopusDeploy/Issues#1506

Sometimes users use the wrong API key with Octo.exe, and can't find projects or environments because the user they are running as doesn't have permission to see them.

After connecting, print the username we connected as
When we can't find a project/environment/etc., in the error message, suggest it might possibly be a permissions issue

Release note: Octo.exe reports authenticated user and has more helpful error messages

### DIFF
--- a/source/OctopusTools/Commands/ApiCommand.cs
+++ b/source/OctopusTools/Commands/ApiCommand.cs
@@ -34,8 +34,8 @@ namespace OctopusTools.Commands
             var options = optionGroups.For("Common options");
             options.Add("server=", "The base URL for your Octopus server - e.g., http://your-octopus/", v => serverBaseUrl = v);
             options.Add("apiKey=", "Your API key. Get this from the user profile page.", v => apiKey = v);
-            options.Add("username=", "[Optional] Username to use when authenticating with the server.", v => username = v);
-            options.Add("password=", "[Optional] Password to use when authenticating with the server.", v => password = v);
+            options.Add("user=", "[Optional] Username to use when authenticating with the server.", v => username = v);
+            options.Add("pass=", "[Optional] Password to use when authenticating with the server.", v => password = v);
             options.Add("configFile=", "[Optional] Text file of default values, with one 'key = value' per line.", v => ReadAdditionalInputsFromConfigurationFile(options, v));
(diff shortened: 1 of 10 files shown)


source/OctopusTools/Commands/ApiCommand.cs
```diff
@@ -34,8 +34,8 @@ protected ApiCommand(IOctopusRepositoryFactory repositoryFactory, ILog log)
             var options = optionGroups.For("Common options");
             options.Add("server=", "The base URL for your Octopus server - e.g., http://your-octopus/", v => serverBaseUrl = v);
             options.Add("apiKey=", "Your API key. Get this from the user profile page.", v => apiKey = v);
-            options.Add("user=", "[Optional] Username to use when authenticating with the server.", v => user = v);
-            options.Add("pass=", "[Optional] Password to use when authenticating with the server.", v => pass = v);
+            options.Add("username=", "[Optional] Username to use when authenticating with the server.", v => username = v);
+            options.Add("password=", "[Optional] Password to use when authenticating with the server.", v => password = v);
             options.Add("configFile=", "[Optional] Text file of default values, with one 'key = value' per line.", v => ReadAdditionalInputsFromConfigurationFile(options, v));
```
michaelnoonan: I think this will be a breaking change to the command line arguments?

source/OctopusTools/Commands/CreateEnvironmentCommand.cs
```diff
@@ -13,7 +13,7 @@ public CreateEnvironmentCommand(IOctopusRepositoryFactory repositoryFactory, ILo
         {
             var options = Options.For("Environment creation");
             options.Add("name=", "The name of the environment", v => EnvironmentName = v);
-            options.Add("ignoreIfExists", "If the project already exists, an error will be returned. Set this flag to ignore the error.", v => IgnoreIfExists = true);
```
michaelnoonan: Ha! Nice pickup!

source/OctopusTools/Commands/ApiCommand.cs
IanMercer: Argh ... resharper ... will fix.

PR #63, opened 2015-11-13
Add flag to enable cancellation of the deployment when timeout is reached

matt-richardson:
We found that the --deploymenttimeout parameter had a bit of a misleading help text, which @Dalmirog updated in #62, based on the [conversation](http://help.octopusdeploy.com/discussions/questions/6140-deploymenttimeout-failing-teamcity-build-but-not-cancelling-octopus-deployment) we had on help.octopusdeploy.com.

We expected that the deployment timeout should cancel the build in octopus, which it did not. This pull request adds a new flag to octo.exe that cancels the deployment(s) if the timeout is reached.

### DIFF
--- a/source/OctopusTools.Tests/Commands/DeployReleaseCommandTestFixture.cs
+++ b/source/OctopusTools.Tests/Commands/DeployReleaseCommandTestFixture.cs
@@ -1,0 +1,78 @@
+﻿using NUnit.Framework;
+using Octopus.Client.Model;
+using OctopusTools.Commands;
+using System;
+using System.Collections.Generic;
+using NSubstitute;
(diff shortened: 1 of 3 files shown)


zentron:
This has been merged into `3.2.2`

zentron:
I've just updated the team city plugin with the latest octo.exe. Thanks @matt-richardson

matt-richardson:
Brilliant. Thanks for that. 

Is that in the latest version available to download (3.2.0.48)? 

(Unfortunately the change log is only available for the main app, not the rest.)

matt-richardson:
Ahh, looks like its in the new download available - 3.2.0.50. 
Thanks!

PR #64, opened 2015-11-18
Move octo.exe into Tools Folder

michaelnoonan:
This is a potential ​*breaking change*​ with the NuGet package we publish to NuGet.org which is to move `octo.exe` into the `tools` folder. This means we don’t need to `DUMMY.txt` hack to make it install (current process), and `octo.exe` will be added to the `$env:PATH` so you can run `octo.exe` commands directly from the ​*PowerShell Manager Console*​.

### DIFF
--- a/source/OctopusTools/DUMMY.txt
+++ b/source/OctopusTools/DUMMY.txt
@@ -1,2 +1,0 @@
-This file is included as a fix to make this work with NuGet 3
-https://github.com/OctopusDeploy/Issues/issues/2100
--- a/source/OctopusTools/OctopusTools.csproj
+++ b/source/OctopusTools/OctopusTools.csproj
@@ -190,7 +190,6 @@
     <None Include="packages.config" />
   </ItemGroup>
   <ItemGroup>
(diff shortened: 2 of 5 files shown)


PashCracken:
Hi,

I'm not sure I like this change. I don't think a NuGet package installed with nuget install should update $env:PATH. NuGet packages should not change the environment. They should only be installed to a local folder.

If you want to install a tool like Octo.exe and have it in the path you use Chocolatey. It's the same NuGet package, but Chocolatey will add the exe file to the path for you.

michaelnoonan:
Hi @PashCracken, updating the `$env:PATH` is actually what `Install-Package` [does by default when there's a tools folder](https://docs.nuget.org/create/creating-and-publishing-a-package#from-a-convention-based-working-directory). It's just that in our testing we found that the `$env:PATH` didn't take effect until after restarting Visual Studio for some reason. I haven't investigated to the bottom of that rat hole and ended up making the choice to spackle over it in a non-breaking way.

At the end of the day I don't think there's any harm in doing this, but I'm happy to be convinced otherwise.

PR #66, opened 2015-12-15
Load additional parameters from a file when --configfile is passed

tothegills:
Did a refactor using `IOctopusFileSystem` instead of `File` so it's a noisy PR.

The important parts are `CreateReleaseCommandFixture.cs` and `CreateRelease.config.txt` which test that parameters can be loaded from a file.

The implementation change is `ApiCommand.cs` line 173 that uses `optionsGroup` instead of just `options`.


tothegills:
Fixes OctopusDeploy/Issues#2233

alfhenrik:
LGTM

PR #69, opened 2016-01-16
Add clean-machines Command

ShawInnes:
New clean-environment Command will remove any machines from an environment which match a specified Status.  The command accepts the following parameters:

--environment="environment name"
--status=Offline

### DIFF
--- a/source/OctopusTools/Commands/CleanEnvironmentCommand.cs
+++ b/source/OctopusTools/Commands/CleanEnvironmentCommand.cs
@@ -30,7 +30,7 @@ namespace OctopusTools.Commands
         protected override void Execute()
         {
             if (string.IsNullOrWhiteSpace(environmentName)) throw new CommandException("Please specify an environment name using the parameter: --environment=XYZ");
-            if (status != MachineModelStatus.Unknown) throw new CommandException("Please specify status using the parameter: --status=Offline");
+            if (status == MachineModelStatus.Unknown) throw new CommandException("Please specify status using the parameter: --status=Offline");
 
             Log.Debug("Loading environment...");
             var environmentResource = Repository.Environments.FindByName(environmentName);
--- a/source/OctopusTools/Commands/CleanEnvironmentCommand.cs
+++ b/source/OctopusTools/Commands/CleanEnvironmentCommand.cs
@@ -47,7 +47,8 @@ namespace OctopusTools.Commands
 
                 foreach (var machine in machines)
                 {
-                    Log.InfoFormat(" - {0} {1} (ID: {2})", machine.Name, machine.Status, machine.Id);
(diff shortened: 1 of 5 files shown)


source/OctopusTools/Commands/CleanEnvironmentCommand.cs
```diff
@@ -0,0 +1,56 @@
...
+            });
+
+            if (machines != null)
+            {
+                Log.Info("Machines: " + machines.Count);
+
+                foreach (var machine in machines)
+                {
+                    Log.InfoFormat("Deleting - {0} {1} (ID: {2})", machine.Name, machine.Status, machine.Id);
+                    Repository.Machines.Delete(machine);
```
PaulStovell: A machine may belong to more than one environment. We should check if that is the case and just remove the machine from the environment instead of deleting it.

ShawInnes: Could a machine be a different status in each environment though?  Maybe my
selection logic is too simplistic.

On 27 January 2016 at 11:05, Paul Stovell <notifications@github.com> wrote:

> In source/OctopusTools/Commands/CleanEnvironmentCommand.cs
> <https://github.com/OctopusDeploy/Octopus-Tools/pull/69#discussion_r50928232>
> :
>
> > +            var environmentResource = Repository.Environments.FindByName(environmentName);
> > +
> > +            Log.Debug("Loading machines...");
> > +            var machines = Repository.Machines.FindMany(x =>
> > +            {
> > +                return x.Status == status && x.EnvironmentIds.Any(environmentId => environmentId == environmentResource.Id);
> > +            });
> > +
> > +            if (machines != null)
> > +            {
> > +                Log.Info("Machines: " + machines.Count);
> > +
> > +                foreach (var machine in machines)
> > +                {
> > +                    Log.InfoFormat("Deleting - {0} {1} (ID: {2})", machine.Name, machine.Status, machine.Id);
> > +                    Repository.Machines.Delete(machine);
>
> A machine may belong to more than one environment. We should check if that
> is the case and just remove the machine from the environment instead of
> deleting it.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/OctopusDeploy/Octopus-Tools/pull/69/files#r50928232>.
>



--

PaulStovell: No, the machine is either offline or online. But say I have:

**Given**: 
Machine Web01 - belongs to Staging and Production, and is Offline

**When:**
I say "clean machines in Staging that are offline"

**Then:**
- The machine would no longer be in Staging
- The machine would still be in Production

Deleting the machine breaks this because it will be deleted from all environments currently.

michaelnoonan:
I've updated this to handle the case where a machine is in multiple environments. In this case we should just remove the machine from the environment.

PR #70, opened 2016-02-17
Feature vstslog

Damovisa:
May need to squash the commits?


michaelnoonan:
This looks good. I like that we're detecting the runtime environment. I also like to see the progression of changes personally.

PR #71, opened 2016-03-07
Fix for Octo.exe to support TLS protocols

thebigmoosey:
Adds support for TLS protocols for Octo.exe to communicate with TLS-only installations of Octopus Server.

Fixes OctopusDeploy/Issues#2414


michaelnoonan:
LGTM - works as advertised. I think this change is quite fine - if someone is happy to use SSLv3, TLS 1.0, 1.1 or 1.2 we shouldn't preclude them. It's good that we've disabled SSLv2, and we should probably consider turning off SSLv3 at some point - though again, that just requires the customer to disable insecure protocols which is better anyhow.

PR #73, opened 2016-03-07
Add --autochannel feature to create-release command

ryangribble:
Hey guys,

This PR adds support to Octo.exe to automatically determine the appropriate channel when creating a release.  This is suggested in [this uservoice item](https://octopusdeploy.uservoice.com/forums/170787-general/suggestions/10776963-automatic-release-creation-to-multiple-channels)

### Todo List
- [x] There is a fair amount of ```log.Debug()``` statements used (see footnote), let me know if this is an issue?  Since we run it from teamcity I always appreciate pretty verbose output so if something goes wrong we can troubleshoot from the teamcity log
- [ ] Adjust the teamcity plugin to cater for the new option (or do we just want users to add ```--autochannel``` in the "additional parameters" input box?

### The approach
I chatted with @michaelnoonan last week and fleshed out our preferred approach for this.  In a nut-shell I check each channel's rules for matches using the octopus api "rule-test" endpoint 
eg: ```api/channels/rule-test?version=1.0.0&versionRange=(1-2]&preReleaseTag=.``` which we thought was nicer than having to implement nuget version range parsing etc in Octo.exe itself

The ```CreateDeployment``` method is kept clean by introducing an ```IChannelResolver``` to either ```ResolveByName()``` or ```ResolveByRule()```.  This class in turn uses a ```IChannelResolverHelper``` to hide away the actual calls to Octopus API and other things... thus allowing unit testing of the channel resolution logic.

### To enable the auto channel selection
Use the ```--autochannel``` flag

eg: ```Octo.exe create-release --server=https://demo.octopusdeploy.com --apikey=API-JBJYBBXNGZLPBBYNHGBPUNAARY --project=Channels.Sample --packageversion=2.3.2-myfeature --version=2.3.2-myfeature --autochannel```

### How the auto channel selection logic works:
- For each channel in the project
 - Generate the release plan for that channel, so we know how many package steps we need to match
 - Check all rules of the channel against the applicable steps, using the existing ```versionResolver``` class to handle versions being set at step, version or global scope
  - If any rule does not match
    - this channel is NOT a candidate
  - If all rules match, but one or more steps are not covered by a rule
    - this channel is NOT a candidate
  - If every step matches it's rule
    - this channel IS a candidate

### Once the candidates are determined we can then select a channel:
- If only 1 candidate
 - this channel is auto selected
- If more than 1 candidate
 - throw an error
- If no candidates 
 - If a default channel exists, select the default
 - If no default channel exists, throw an error

### Testing
Unit tests have been added to assert the behaviour of the ```ChannelResolver.ResolveByName()``` and ```ChannelResolver.ResolveByRule()``` whilst mocking out the ```IChannelResolverHelper``` interface to keep details of Octopus calls etc away
I've also done runtime testing against an octopus instance covering various scenarios.  I tested some complicated type setups, but most importantly this works on what I consider to be a sane/typical setup with only a handful of deployment steps per project, package versions set by ```--packageversion``` rather than varying from eachother, and fairly straight forward channels/rules

### Sample Output
```
Octopus: Handshake successful. Octopus version: 3.3.1; API version: 3.0.0
Octopus: Authenticated as: ryan.gribble <API-xxxxxxxxxxxxxxxxxxxxxxxxxx> 
Octopus: Finding project: ChannelTesting
Octopus: Calculating channel automatically
Octopus: Channel "Beta": Processing 2 rules
Octopus: Channel "Beta": 2 steps need to match channel rules
Octopus: Channel "Beta": Checking rule VersionRange:2.0 Tag:
Octopus: Channel "Beta": Checking step: Deploy WebAPI Package
Octopus: Channel "Beta": Resolved version 2.1.0-test for step: Deploy WebAPI Package
Octopus: Channel "Beta": Calling Octopus API to test rule => api/channels/rule-test?version=2.1.0-test&versionRange=2.0&preReleaseTag=
Octopus: Channel "Beta": API Response: SatisfiesVersionRang=True SatisfiesPreReleaseTag=True
Octopus: Channel "Beta": Checking rule VersionRange:2.1 Tag:
Octopus: Channel "Beta": Checking step: Deploy new package (Beta only)
Octopus: Channel "Beta": Resolved version 2.1.0-test for step: Deploy new package (Beta only)
Octopus: Channel "Beta": Calling Octopus API to test rule => api/channels/rule-test?version=2.1.0-test&versionRange=2.1&preReleaseTag=
Octopus: Channel "Beta": API Response: SatisfiesVersionRang=True SatisfiesPreReleaseTag=True
Octopus: Channel "Beta": Matched 2 and failed 0 of 2 required steps
Octopus: Channel "Beta": Matches all steps and is a viable AutoChannel candidate
Octopus: Channel "Default": Skipping due to 0 rules
Octopus: Channel "Stable": Processing 1 rules
Octopus: Channel "Stable": 1 steps need to match channel rules
Octopus: Channel "Stable": Checking rule VersionRange:[2,3) Tag:^$
Octopus: Channel "Stable": Checking step: Deploy WebAPI Package
Octopus: Channel "Stable": Resolved version 2.1.0-test for step: Deploy WebAPI Package
Octopus: Channel "Stable": Calling Octopus API to test rule => api/channels/rule-test?version=2.1.0-test&versionRange=[2,3)&preReleaseTag=^$
Octopus: Channel "Stable": API Response: SatisfiesVersionRang=True SatisfiesPreReleaseTag=False
Octopus: Channel "Stable": Matched 0 and failed 1 of 1 required steps
Octopus: Channel "Stable": Not a viable AutoChannel candidate due to some steps failing version rules
Octopus: Channel "Beta": Has been automatically selected
```


michaelnoonan:
This is looking really good. I've tested it out on a few scenarios that worked nicely. Just gathering some thoughts after running through the PR this morning:

- The helper, injection, `SetContext()` does feel a bit weird, and I know we use a similar pattern elsewhere. I would have liked to see it more become a case of each method you call on the Helper taking the arguments it needs at the time of the call, rather than setting and caching the `Project`, for example.
- Logging: I would like to see it form up as two different styles of log. A more terse log when everything worked as expected (a Channel was auto-selected), showing the Channels that were considered, and a summary as to why they were excluded. A more verbose log that's dumped out in the case where a Channel wasn't auto-selected, showing the entire process. This way we have the information that makes the most sense for both contexts. (I like the verbose logging for supporting this feature)
- I hit a scenario where two channels ended up matching, one of them the default, but the process failed `System.Exception: Could not determine channel automatically as 2 channels (1.x Hotfix,1.x Stable) matched all steps` I think in this case it would make sense to choose the default channel.
- It would be good to add a `dryRun` / `whatIf` so people can test it out without actually creating a release.

Looking like you're on to a winning idea though!

credible:
Thumbs up on the overall strategy.

Zhaph:
Looks like a great move - thanks.

@michaelnoonan Not sure about "if it matches two and one is the default choose default" - I think the argument could be made either way (choose more specific over default might be appropriate in some cases).

If we're only creating a release, rather than actually deploying direct from TeamCity it might be better to warn, but not fail the build in that case - if my Channels are ambiguous I'd probably still like to create a manual release in Octopus which I can't do if the build has failed and the Artifacts aren't published...

ryangribble:
> The helper, injection, SetContext() does feel a bit weird, and I know we use a similar pattern elsewhere. I would have liked to see it more become a case of each method you call on the Helper taking the arguments it needs at the time of the call, rather than setting and caching the Project, for example.

Yeah if you look at the commit history this is how I had it at first but it felt very messy/clunky so I reworked it to what's in the PR.  I dont like the fact that the helper set context is called from outside the ChannelResolver but I also really didnt like passing ```repository```, ```project```, ```versionResolver```, ```deploymentprocess``` etc into pretty much EVERY call, or making multiple API calls to load the same thing (eg deployment process) whenever it's needed)
I was hoping someone might have a cool suggestion for a better autofac pattern for allowing this kind of context based helper class.
Did you want me to roll back to this previous incarnation or can we leave the magic helper context in place?

> Logging: I would like to see it form up as two different styles of log

Good suggestion, I'll rework the logging to output just the info level statements by default, and collect up the rest into a list of verbose output that can be dumped out if a problem occured

>  I think in this case it would make sense to choose the default channel

> Not sure about "if it matches two and one is the default choose default" - I think the argument could be made either way (choose more specific over default might be appropriate in some cases).

Yeah this is precisely why I decided to fail at this point as it doesnt seem possible to make a decision that will be correct for everyone.  In my thinking I would never expect to have more than 1 channel be allowed to match - eg in your case @michaelnoonan how could a release be a 1.x hotfix AND a 1.x stable?
Personally I'd probably vote to leave things as I've implemented - at least it's a simple/obvious rule set, and users can configure their channels so that these multi match situations dont occur.
However, if it's a deal breaker for people (id love to know the use cases) we could introduce a parameter (or just have values on the ```--autoChannel``` parameter) to specify the desired default treatment in match conflict scenario ```default``` to choose default (assuming it is one of the options) ```specific``` to go with the non default one (but really how is it any more specific than default consdering to get to this point they both had to match all steps?) and ```none``` to throw the error (current behaviour)

> If we're only creating a release, rather than actually deploying direct from TeamCity it might be better to warn, but not fail the build in that case - if my Channels are ambiguous I'd probably still like to create a manual release in Octopus which I can't do if the build has failed and the Artifacts aren't published...

Personally Im not that squeamish about failing a build, cos it's a sure fire way to get the team/person involved to sort out the issue (ie channel rules in the octopus project arent set appropriately, or some disconnect between the version numbering in teamcity and the intention of the octopus project).  If you let something like this "silently" fail (ie convert to a warning) then you potentially let these problems go unaddressed.  A bit like if you have a step that runs unit tests but you tell it not to fail the build if the tests fail.  Build pipelines are cheap... just fix up the problem and trigger another build.  Should you really be publishing/deploying anything that didnt match the "rules" that were setup?

There are also lots of scenarios where this could be dangerous - eg as you mention things like automatic deployment (either through octopus lifecycle or by subsequent build steps or configurations).  There can be some pretty fundamental differences in octopus deployment processes and variables, based on channel and given that this ```autoChannel``` feature is already something people would opt in to, I dont think it's unreasonable to impose an approach of "the rules are the rules" when they do opt-in.

> It would be good to add a dryRun / whatIf so people can test it out without actually creating a release.

I like this too.  I was just specifying an existing version number without the --force option so it would eventually fail to create the release, but get through all the channel resolving logic first :laughing: 

So would this ```--dryRun``` flag be global to the whole ```createrelease``` command, ie do everything right up to the actual release creation but skip over that part?  And what about the inline ```DeployRelease()``` option at the end... should the dryRun cascade into that or it's ok to skip that (since we didnt actually create a release in the first place).

ryangribble:
Sorry for the slackness in getting this over the line!  :grinning:

I've made a couple of tweaks

- Implemented the discussed verbose logging (collect up trace info and dump it at the end only if there was a problem)
- Implemented a "WhatIf" flag to do a dry run
I havent changed any of the autofac stuff sorry, hope that's OK

Also note i tweaked a couple of log messages which you may or may not be happy with the wording

And one final thing - the [IgnoreIfAlreadyExists](https://github.com/OctopusDeploy/Octo.exe/blob/master/source/Octopus.Cli/Commands/CreateReleaseCommand.cs#L155) logic appears to be backwards (I dug down and it seems it has been this way since 2014???) because it only pre-emptively checks if a release exists when that Ignore flag is TRUE but I'm pretty sure it should be doing it when flag is FALSE.  But given this has been like this for a couple of years I didnt want to make that change in this PR so instead am just mentioning it to you!

michaelnoonan:
@ryangribble you're a champion my friend. I've worked today on the code from your PR. I've had to change a bunch of it but I think the concept is really solid, and the end result, I think, is looking really good. Kudos.

The main changes are:

1. Make building a `ReleasePlan` more consistent (there were going to be issues with calculating the Version for each Step in different ways...)
1. Build a `ReleasePlan` per `Channel` and pick the most viable (using the rules we'd already discussed)
1. Use the Hypermedia links for Version Rules test URL (also fixed a URL encoding issue with RegEx - so many `/`!)

I've created a new PR for this since I can't rebase this PR directly.

Continued in https://github.com/OctopusDeploy/Octo.exe/pull/75

PR #74, opened 2016-03-30
Bug custom expression feed

thebigmoosey:
Fix to allow custom expressions in Nuget feeds for projects.

The exporter and importer were attempting to verify all feeds against the API, but there are times when the project has custom expressions for the feed ids using Octostache variables. So this fix makes sure we only check feeds against the API if they are _real_ DB records, and not a custom expression feed.

Fixes OctopusDeploy/Issues#2432

### DIFF
--- a/source/Octopus.Cli/Util/FeedResourceCustomExpressionHelper.cs
+++ b/source/Octopus.Cli/Util/FeedResourceCustomExpressionHelper.cs
@@ -3,7 +3,7 @@
 namespace Octopus.Cli.Util
 {
     /// <summary>
-    /// Helps with situation where feeds use custom expressions Eg. #{MyCustomFeedURL}
+    /// Helps with situations where feeds use custom expressions Eg. #{MyCustomFeedURL}
     /// </summary>
     public static class FeedResourceCustomExpressionHelper
     {
--- a/source/Octopus.Cli/Util/FeedResourceCustomExpressionHelper.cs
+++ b/source/Octopus.Cli/Util/FeedResourceCustomExpressionHelper.cs
@@ -1,0 +1,39 @@
+﻿using Octopus.Client.Model;
+
+namespace Octopus.Cli.Util
+{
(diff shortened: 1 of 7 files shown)


source/Octopus.Cli/Util/FeedResourceCustomExpressionHelper.cs
```diff
@@ -0,0 +1,39 @@
...
+        /// Feeds may have custom expressions as their Id, which may contain Octostache variable syntax #{MyCustomFeedURL}.
+        /// If you pass a custom expression Id like this to the API, it will resolve to /api/feeds/, return all the feeds, then
+        /// attempt to cast that response to a FeedResource object and you'll end up getting an empty FeedResource object instead 
+        /// of null. This method helps you detect valid repository feed objects before running into this confusing API scenario.
+        /// </summary>
+        /// <param name="id"></param>
+        /// <returns></returns>
+        public static bool IsValidRepositoryId(string id)
+        {
+            if (id.ToLowerInvariant().StartsWith("feeds-"))
```
michaelnoonan: I like that you're detecting the positive case!

Try `id.StartsWith("feeds-", StringComparison.OrdinalIgnoreCase)` which is what we commonly use, rather than converting the entire string to lower case.

michaelnoonan:
I like the approach, especially to detecting that it's a real feed by `Id.StartsWith("feeds-")`.

One thought I have is rather than `ValidFeedId` consider something that distinguishes between a `RealFeedId` and a `Fake` or `ExpressionFeedId`. I found myself getting confused at first by the method names, and had to read the code to be sure what was happening.

Some other thoughts inline.

thebigmoosey:
The exporter and importer were attempting to verify all feeds against the API, but there are times when the project has custom expressions for the feed ids using Octostache variables. So this fix makes sure we only check feeds against the API if they are _real_ DB records, and not a custom expression feed.

PR #77, opened 2016-04-29
Add "create-channel" command.

uglybugger:
# Rationale
The rationale behind this feature is to allow branch builds from TeamCity or other build servers to have a per-branch channel created for them.

`octo.exe create-channel --server=XYZ --apiKey=API-XYZ --channel="My New Channel" --project="My Project" --lifecycle=Default --update-if-exists`

# Use case
Use case: a team using short-lived feature branches wants to create a different IIS binding for each branch. The current best solution appears to be to parse a release version number or a package name and extract a branch name into an output variable.

This solution would allow a much cleaner creation of a channel for any given branch. I discussed this with @michaelnoonan a while ago and offered to send a pull request if the Octopus team didn't get around to building it first. Here it is :)

# Implementation
This is a fairly straight-forward wrapper over the Octopus REST API.

There's one small feature that might be noteworthy: the `--update-if-exists` parameter, which allows "create or update" behaviour.

### DIFF
--- a/source/Octopus.Cli/Commands/CreateChannelCommand.cs
+++ b/source/Octopus.Cli/Commands/CreateChannelCommand.cs
@@ -48,13 +48,13 @@ namespace Octopus.Cli.Commands
             else
             {
                 Log.DebugFormat("Loading lifecycle {0}", lifecycleName);
-                var lifecycle = Repository.Lifecycles.FindOne(l => l.Name == lifecycleName);
+                var lifecycle = Repository.Lifecycles.FindOne(l => string.Compare(l.Name, lifecycleName, StringComparison.OrdinalIgnoreCase) == 0);
                 if (lifecycle == null) throw new CouldNotFindException("lifecycle named", lifecycleName);
                 lifecycleId = lifecycle.Id;
             }
 
             var channelsForThisProject = Repository.Client.List<ChannelResource>(project.Links["Channels"]);
-            var channel = channelsForThisProject.Items.FirstOrDefault(ch => ch.Name == channelName);
+            var channel = channelsForThisProject.Items.FirstOrDefault(ch => string.Compare(ch.Name, channelName, StringComparison.OrdinalIgnoreCase) == 0);
 
             if (channel == null)
             {
--- a/source/Octopus.Cli/Commands/CreateChannelCommand.cs
+++ b/source/Octopus.Cli/Commands/CreateChannelCommand.cs
@@ -35,15 +35,23 @@ namespace Octopus.Cli.Commands
         {
             if (string.IsNullOrWhiteSpace(projectName)) throw new CommandException("Please specify a project using the parameter: --project=ProjectXYZ");
             if (string.IsNullOrWhiteSpace(channelName)) throw new CommandException("Please specify a channel name using the parameter: --channel=ChannelXYZ");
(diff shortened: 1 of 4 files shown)


michaelnoonan:
Nice. I can see how this will become more and more useful over time also. As part of Octopus 3.4 we are rebuilding the dashboard views so you can pivot on different axes, like showing environment/channel which would show you the latest release in each environment per channel.

After you're done with that feature branch it would be nice to easily delete the feature channel, and have all the releases get cleaned up... Food for thought.

One comment inline. :)

source/Octopus.Cli/Commands/CreateChannelCommand.cs
```diff
@@ -0,0 +1,81 @@
...
+        string lifecycleName;
+        string channelDescription;
+        bool updateIfExists;
+        bool? makeDefaultChannel;
+
+        protected override void Execute()
+        {
+            if (string.IsNullOrWhiteSpace(projectName)) throw new CommandException("Please specify a project using the parameter: --project=ProjectXYZ");
+            if (string.IsNullOrWhiteSpace(channelName)) throw new CommandException("Please specify a channel name using the parameter: --channel=ChannelXYZ");
+            if (string.IsNullOrWhiteSpace(lifecycleName)) throw new CommandException("Please specify a lifecycle name using the parameter: --lifecycle=LifecycleXYZ");
```
michaelnoonan: A `null` lifecycle is allowed, which is interpreted as "Use the project's default lifecycle". I think for consistency `octo.exe` should allow the same behaviour as the UI.

uglybugger:
I've updated the pull request to make the lifecycle name non-mandatory as per the API. It's also updated to allow for case-insensitive matches of channel and lifecycle names (again, as per the API).

yngvebn:
Not sure if commenting on closed issues is ok, or not, but here goes nothing :) 
This feature would be extremely useful for us if it was also possible to supply version-range/tag-prefix to the create-channel, as we're attempting to get automatic release-creating from TeamCity going. As the build will publish package(s) with version 1.2.3-MYFEATURE1, the newly created channel should have a tag-regex of "MYFEATURE1", as those packages with that branch are the only ones that are valid for this release.

Maybe we're going about this the wrong way, so if you guys have any feedback or suggestions to better solutions to this, I would be extremely grateful :)

michaelnoonan:
Hi @yngvebn. I think it's better to talk than stay silent! https://octopus.com/support is a really good way to get help too. :)

Creating a new Channel complete with package Version Rule(s) would be great, especially now we automatically choose the best Channel for a Release based on the Packages involved... but I can see the syntax becoming tricky. Totally shooting from the hip for this example:

`octo.exe create-channel --server=XYZ --apiKey=API-XYZ --channel="MYFEATURE1" --project="My Project" --lifecycle="Feature Branch Lifecycle" --versionRule="Package Step 1":"[1.0,2.0)":"MYFEATURE1" --versionRule="Package Step 2":"[1.0,2.0)":"MYFEATURE1" --update-if-exists`

Regardless of the exact syntax, we would need to parse everything after the `--versionRule=` into the Version Range and Tag Regex parts. :)

What syntax would make the most sense to you? @uglybugger would be interested for you to weigh in also. :)

Also for context, we do have [this suggestion](https://octopusdeploy.uservoice.com/forums/170787-general/suggestions/10920369-allow-creating-a-channel-version-rule-that-applies) floating around also.

PR #78, opened 2016-06-05
Issue releaseversionfrompackage

slewis74:
Updated the parameter parsing, to ensure that packageversion only contains a semantic version.  Outputs a message to the user to direct them to the package parameter, if it looks like they're trying to use StepName:PackageVersion

Fixes OctopusDeploy/Issues#2533


droyad:
Should the validation go with the other version parameter validation in `PackageVersionResolver`?

slewis74:
Yep.  I had it in there and then moved it because I wasn't convinced that was the correct behavior.  Looking at it again, that's where it should be.

PR #80, opened 2016-08-11
Replace log4net with serilog in preparation for .net core

markryd:
I've moved to ILogger rather than shimming ILog because there wasn't much to it and I figured I might as well.

log4net aren't in a hurry to move to .net core (there is a PR that has been drifting along) and this is pretty much the only place we use log4net, and we're only using it to write to the console, so replace with serilog.


michaelnoonan:
Looks like a no-brainer from my point of view.

PR #81, opened 2016-09-13
Updated Readme

droyad:


readme.md
```diff
@@ -1,12 +1,30 @@
...
+- [Octo.exe][4]
+- [Octopus.Client][5]
 
-Documentation can be found in the [Octopus Deploy documentation](http://docs.octopusdeploy.com/display/OD/Octo.exe+Command+Line). 
+## Issues
+Please see [Contributing](CONTRIBUTING.md)
 
-**Please raise and track issues for this project [here](https://github.com/OctopusDeploy/Issues/issues/).**
+## Compatibility
+Version 3.3 and later of the clients are backwards compatible with Octopus 3.0 or later. Obviously the available features and APIs are limited to those supported by server.
```
michaelnoonan: Hmmm, I find this a bit confusing, because I thought 3.2 of the client was also backwards compatible etc...

michaelnoonan:
I like it. Makes clear some of our unwritten intentions.

readme.md
droyad: Ah true, I just picked one I knew. Any idea how far back?

michaelnoonan: We shipped Client 3.0 with Octopus 3.0, and we've tried really hard not to break compatibility... Except in Octopus 3.3 we did change `SensitiveProperties`...

Perhaps it's safest to say something like:

Octopus Server 3.0-3.2 use Octopus Client 3.0-3.2 (newer is better)
Octopus Server 3.3 use Octopus Client 3.3 or newer [release notes](https://octopus.com/downloads/3.3.0)
Octopus Server 3.4 use Octopus Client 3.4 or newer [release notes](https://octopus.com/downloads/3.4.0)

PR #90, opened 2016-09-29
Add create and modify behavior to user roles repository

williamlord:
I'm extending https://github.com/Suremaker/OctopusProjectBuilder to support user roles and teams, and I'm currently having to drop down to the client for create and modification.

I'd like to be able to do these operations through the repository instead.

The only role that can't be modified is the built-in system administrator role, this is handled by the API.

I've added signatures to public surface area test, however it's failing on master head relating to subscriptions.


PaulStovell:
Thanks for this pull request @williamlord!

williamlord:
Thanks for merging.

PR #106, opened 2016-12-08
Correction for Gathering ProjectGroup Projects

ChadRoesler:
The Link that is called on the projectGroupResource does not eixist.

The actual link it should be pulling is Projects.
This is directly pulled from the API in 3.6.0

``` JSON
{
  "Id": "ProjectGroups-1",
  "Name": "Active Clients",
  "Description": "Active clients.",
  "EnvironmentIds": [],
  "RetentionPolicyId": null,
  "Links": {
    "Self": "/api/projectgroups/ProjectGroups-1",
    "Projects": "/api/projectgroups/ProjectGroups-1/projects"
  }
}  
```

### DIFF
--- a/source/Octopus.Client/OctopusRepository.cs
+++ b/source/Octopus.Client/OctopusRepository.cs
@@ -1083,7 +1083,7 @@ namespace Octopus.Client
             {
                 var resources = new List<ProjectResource>();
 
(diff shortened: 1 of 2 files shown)


droyad:
@ChadRoesler Thank you for the fix, it has been published as version 4.2.4

PR #107, opened 2016-12-14
Project and Task Enhancements

ChadRoesler:
Added method for gathering all Releases related to project, rather than having to page through them.

Added method for gathering the Tasks that a Task currently is queued behind.


droyad:
Hi @ChadRoesler thank you for your contribution. It has been released as v4.3.0

PR #123, opened 2017-02-09
Support determining package name and version from zip file name

lahma:
Wrote test case that shows that now zip file version are also handled. Added separate logic for zip file names that uses only file name to determine version, without opening the actual file.

Fixes https://github.com/OctopusDeploy/Issues/issues/3148


droyad:
@lahma Thank you for the contribution, I've just got to check whether we support `<id>-<version>.zip` as well as `<id>.<version>.zip`

PR #124, opened 2017-02-10
Add zip version parsing to release creation process

droyad:
Resolves https://github.com/OctopusDeploy/Issues/issues/3148

In the server we support `<id>-<version>.zip` as well as `<id>.<version>.zip` in different places, but not at the same time. Need to check which is right (or both).

### DIFF
--- a/source/Octo/Commands/PackageVersionResolver.cs
+++ b/source/Octo/Commands/PackageVersionResolver.cs
@@ -17,14 +17,12 @@ namespace Octopus.Cli.Commands
 {
     public class PackageVersionResolver : IPackageVersionResolver
     {
-        static readonly string[] SupportedZipExtensions = { ".zip", ".tgz", ".tar.gz", ".tar.Z", ".tar.bz2", ".tar.bz", ".tbz", ".tar" };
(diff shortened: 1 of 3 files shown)


source/Octo/Commands/PackageVersionResolver.cs
```diff
@@ -1,29 +1,38 @@
...
 
 namespace Octopus.Cli.Commands
 {
     public class PackageVersionResolver : IPackageVersionResolver
     {
         readonly Serilog.ILogger log;
+        private readonly IOctopusFileSystem fileSystem;
         readonly IDictionary<string, string> stepNameToVersion = new Dictionary<string, string>(StringComparer.CurrentCultureIgnoreCase);
         string defaultVersion;
+        readonly Regex zipNameRegex = new Regex(@"(?<Name>.+?)\.(?<Version>\d+(\s*\.\s*\d+){0,3})(?<Release>-[a-z][0-9a-z-]*)?\.zip$", RegexOptions.Compiled | RegexOptions.IgnoreCase | RegexOptions.ExplicitCapture);
```
markryd: Not used, delete.

PR #133, opened 2017-02-22
Add support for anti-forgery token using cookie-to-header pattern

michaelnoonan:
Based on this recommendation: https://en.wikipedia.org/wiki/Cross-site_request_forgery#Cookie-to-Header_Token

See related Octopus PR: https://github.com/OctopusDeploy/OctopusDeploy/pull/847

> If there is an anti-forgery cookie in the response, copy its value to the correct header for any outbound requests

# Compatibility concerns

> Customers will only need a newer Client version if they upgrade their Octopus Server and are using Username Password credentials

1. Customers using API Keys will not be affected - API Key authenticated requests are not subject to anti-forgery token validation.
2. Older servers don't send the anti-forgery cookie so it can't/won't be copied to any outgoing requests
3. Newer servers will require anti-forgery header if the client uses an auth cookie session, and will make recommendations if the anti-forgery header is missing

I think this approach should cover 99% of cases:
1. Most customers should be using API Keys for automation - no change
2. Those customers who use username/password just need to grab the latest client
3. Those using raw HTTP will need to implement their own cookie-to-header copy

### DIFF
--- a/source/Octopus.Client/OctopusClient.cs
+++ b/source/Octopus.Client/OctopusClient.cs
@@ -464,7 +464,7 @@ namespace Octopus.Client
                 .SingleOrDefault(c => c.Name.StartsWith(ApiConstants.AntiforgeryTokenCookiePrefix));
             if (antiforgeryCookie != null)
             {
-                //webRequest.Headers[ApiConstants.AntiforgeryTokenHttpHeaderName] = antiforgeryCookie.Value;
+                webRequest.Headers[ApiConstants.AntiforgeryTokenHttpHeaderName] = antiforgeryCookie.Value;
             }
 
             var requestHandler = SendingOctopusRequest;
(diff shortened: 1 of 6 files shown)


source/Octopus.Client/OctopusClient.cs
```diff
@@ -447,6 +459,14 @@ OctopusResponse<TResponseResource> DispatchRequest<TResponseResource>(OctopusReq
                 webRequest.Method = "POST";
             }
 
+            var antiforgeryCookie = cookieContainer.GetCookies(cookieOriginUri)
+                .Cast<Cookie>()
+                .SingleOrDefault(c => c.Name.StartsWith(ApiConstants.AntiforgeryTokenCookiePrefix));
+            if (antiforgeryCookie != null)
+            {
+                //webRequest.Headers[ApiConstants.AntiforgeryTokenHttpHeaderName] = antiforgeryCookie.Value;
```
zentron: should we assigning stuff here like above in async version? Maybe in some common code?

michaelnoonan:
Oh my! When test code gets pushed! Good pickup, I'll make a shared implementation!

-----Original Message-----

michaelnoonan:
I've added some tests for both Async and Sync clients. There was very little code which could be shared due to the different implementations. I think the tests should prevent us making mistakes now since it proves the end-to-end behaviour based on the expected Cookie and Header names.

PR #144, opened 2017-03-10
pack: --verbose flag for files in package logging

NickJosevski:
pack command takes --verbose, will log out more info to start with all
the files added to ZIP and NuPkg

Fixes: https://github.com/OctopusDeploy/Issues/issues/3171

Taks:
 - [x] octo.exe support of --versbose flag

Will come later:
 - VSTS extension supporting changes
 - Team City extension supporting changes (java)

### DIFF
--- a/source/Octo/Commands/NuGetPackageBuilder.cs
+++ b/source/Octo/Commands/NuGetPackageBuilder.cs
@@ -18,12 +18,18 @@ namespace Octopus.Cli.Commands
             this.log = log;
         }
 
-        public void BuildPackage(string basePath, IList<string> includes, ManifestMetadata metadata, string outFolder, bool overwrite)
+        public void BuildPackage(string basePath, IList<string> includes, ManifestMetadata metadata, string outFolder, bool overwrite, bool verboseInfo)
         {
             var nugetPkgBuilder = new PackageBuilder();
 
-            nugetPkgBuilder.PopulateFiles(basePath, includes.Select(i => new ManifestFile { Source = i }));
+            var manifestFiles = includes.Select(i => new ManifestFile {Source = i}).ToList();
+            nugetPkgBuilder.PopulateFiles(basePath, manifestFiles);
             nugetPkgBuilder.Populate(metadata);
+
+            if (verboseInfo)
+            {
+                nugetPkgBuilder.Files.ToList().ForEach(f => log.Information("Adding file: {Path}", f.Path));
+            }
 
             var filename = metadata.Id + "." + metadata.Version + ".nupkg";
             var output = Path.Combine(outFolder, filename);
--- a/source/Octo/Commands/NuGetPackageBuilder.cs
+++ b/source/Octo/Commands/NuGetPackageBuilder.cs
@@ -28,7 +28,10 @@ namespace Octopus.Cli.Commands
 
             if (verboseInfo)
             {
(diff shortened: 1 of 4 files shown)


source/Octo/Commands/PackCommand.cs
```diff
@@ -127,9 +129,10 @@ public Task Execute(string[] commandLineArguments)
                 if (!string.IsNullOrWhiteSpace(title))
                     metadata.Title = title;
 
-                log.Information("Packing {PackageId:l} version {Version}...", id, version);
+                log.Information($"{(verbose ? "Verbose logging" : "")}");
```
markryd: I reckon just do an if here rather than log an empty line when not verbose?

source/Octo/Commands/NuGetPackageBuilder.cs
```diff
@@ -18,13 +18,19 @@ public NuGetPackageBuilder(IOctopusFileSystem fileSystem, Serilog.ILogger log)
...
             var nugetPkgBuilder = new PackageBuilder();
 
-            nugetPkgBuilder.PopulateFiles(basePath, includes.Select(i => new ManifestFile { Source = i }));
+            var manifestFiles = includes.Select(i => new ManifestFile {Source = i}).ToList();
+            nugetPkgBuilder.PopulateFiles(basePath, manifestFiles);
             nugetPkgBuilder.Populate(metadata);
 
+            if (verboseInfo)
+            {
+                nugetPkgBuilder.Files.ToList().ForEach(f => log.Information("Adding file: {Path}", f.Path));
```
markryd: Sorry, pet hate of mine is making a list just so you get access to ForEach on it. Just use a `foreach` loop, or if you really want `ForEach` make an extension method on IEnumerable or something.

source/Octo/Commands/ZipPackageBuilder.cs
```diff
@@ -47,7 +47,11 @@ public void BuildPackage(string basePath, IList<string> includes, ManifestMetada
 
                         var relativePath = UseCrossPlatformDirectorySeparator(
                             fullFilePath.Substring(basePathLength).TrimStart(Path.DirectorySeparatorChar, Path.AltDirectorySeparatorChar));
-                        log.Debug("Adding file: {Path}", relativePath);
+
+                        if (verboseInfo)
+                            log.Information("Adding file: {Path}", relativePath);
```
droyad: Does Log have a `Write(LogLevel, string, params string[])` override you can use instead? (The message is exactly the same)

source/Octo/Commands/NuGetPackageBuilder.cs
droyad: Should this be `Added` instead of `Adding`?

source/Octo/Commands/ZipPackageBuilder.cs
NickJosevski: Yes it does can tidy that up

source/Octo/Commands/PackCommand.cs
NickJosevski: Yes that's better, not sure why I decided to log an empty line there (I may have had it saying 'non verbose' while testing)...

source/Octo/Commands/NuGetPackageBuilder.cs
NickJosevski: Yes will switch to 'Added' for the lines where it's doing it, here and in the Zip Package

Yea will do a loop instead :)

PR #145, opened 2017-03-13
Create user

michaelnoonan:
The ability to create users was missing from the basic repository (by omission I think), and have added some shortcut methods for creating valid user accounts for people and services.

Tested against Octopus 3.11.x but should be backwards and forwards compatible since there are no resource changes, and it's an additive API change.


pawelpabich:
Looks great

PR #146, opened 2017-03-13
Isolate anti-forgery token by InstallationId

michaelnoonan:
In the rare-ish case where customers have multiple Octopus installations on the same domain (it does happen, quite a lot in our test environments too) we will end up with multiple cookies in the CookieCollection for that domain, and we were looking for the first match based on the common prefix. Quite often this would end up with the wrong anti-forgery cookie for that installation and fail validation. We should use the InstallationId to partition and find the correct anti-forgery cookie.

image

This should be both backwards and forwards compatible. We haven't released any versions of Octopus Server with CSRF support yet, so there also shouldn't be any compatibility concerns there.

See related commit for Server and JS: https://github.com/OctopusDeploy/OctopusDeploy/pull/847/commits/67bc3f64f8e5ac16e3b307392a8849bb82ef212d


zentron:
Cant think of any issues that this might cause . LGTM

PR #147, opened 2017-03-19
Update actions that are based on custom step templates

pawelpabich:


zentron:
Looks good. The new error property could come in handy

PR #150, opened 2017-03-21
Fix unix octo start script so that it

roederja2:
- can be called via symlinks from any location
- works in docker
- passes on the arguments


droyad:
@roederja2 Thank you for that! The change will be out shortly as `4.13.6`

PR #154, opened 2017-03-31
Project dash

zentron:
Just added property.. The config change was just a re-sort


zentron:
[Deployed version highlighting not correct for channels on project dashboard](https://trello.com/c/na0B4EQ0/33-deployed-version-highlighting-not-correct-for-channels-on-project-dashboard)

PR #168, opened 2017-05-24
Fix for PackageId-based version overrides

bartoncasey:
The current --package argument handling doesn't work as described in https://octopus.com/docs/api-and-integration/octo.exe-command-line/creating-releases. 

Specifically, if you
- provide a default `--packageversion=1.1.1` argument
- do NOT  provide a `--package=stepname:1.2.3` argument 
- provide a `--package=my.package.id:1.2.3` argument

In this case the default will be preferred over the packageid override.

This fix changes the fallback to stepname -> packageid -> default.


droyad:
@bartoncasey Wow, I can't believe it's been broken for like ever. Thanks for the PR, I'll merge as soon as the CLA is signed.

droyad:
Fixes #169

PR #172, opened 2017-06-05
Add overloads that take interfaces

matt-richardson:
So we can mock a bit easier


source/Octopus.Client/Operations/IRegisterMachineOperation.cs
```diff
@@ -82,6 +82,12 @@ public interface IRegisterMachineOperation
         /// </summary>
         /// <param name="repository">The Octopus Deploy repository.</param>
         void Execute(OctopusRepository repository);
+
```
michaelnoonan: Is the duplication actually required?

PR #174, opened 2017-06-12
TenantedDeploymentParticipation property

MJRichardson:
Added TenantedDeploymentParticipation property to Machine, Account, and Certificate resources.

Closes https://github.com/OctopusDeploy/OctopusClients/issues/184
Related to https://github.com/OctopusDeploy/Issues/issues/2722
Dependent on https://github.com/OctopusDeploy/OctopusDeploy/pull/1146


MJRichardson:
[Enable untenanted projects to be deployed to tenanted deployment targets](https://trello.com/c/tpc4wCYz/37-enable-untenanted-projects-to-be-deployed-to-tenanted-deployment-targets)

source/Octopus.Client/Editors/Async/MachineEditor.cs
```diff
@@ -54,20 +55,29 @@ public MachineEditor(IMachineRepository repository)
             EnvironmentResource[] environments,
             string[] roles,
             TenantResource[] tenants,
-            TagResource[] tenantTags)
+            TagResource[] tenantTags, 
+            TenantedDeploymentParticipation? tenantedDeploymentParticipation = null)
```
zentron: if we _dont_ allow it to be nullable, then is that a problem? Wont old servers just ignore the property.

MJRichardson: My thinking was that making it nullable meant existing code won't break.  Otherwise it will cause a compile-error when upgrading.

PR #187, opened 2017-07-07
Added support to filter/search project releases

robpearson:
Project releases can now be searched via a version filter.  This is related to OctopusDeploy/Issues#3533

Tests:
- Updated existing tests only.  Adding API test in Octopus E2E Tests for better coverage.


robpearson:
@michaelnoonan Yeah, I removed the xml help snippet as it wasn't populated.  Good point regarding the descriptiveness of the new param.  I'll rename and revise.  Thanks for the review! :)

PR #189, opened 2017-08-07
URL encoding for extended ASCII characters was producing the wrong re…

### LINKED ISSUE #3718: Octo.exe doesn't handle the character ä for project names
Octopus Server: `3.16.0`

This used to work in `Octo.exe` `4.2.4` but got broken somewhere in the way to `4.22.0`

octoexespecialcharacter

Source: http://help.octopusdeploy.com/discussions/problems/56562

Definitely worth checking if the same will happen for Channel name, Environment Name, etc

benPearce1:
Fixes https://github.com/OctopusDeploy/Issues/issues/3718 
Added base unit tests around UrlTemplate

### DIFF
--- a/source/Octopus.Client.Tests/UrlTemplateTests.cs
+++ b/source/Octopus.Client.Tests/UrlTemplateTests.cs
@@ -1,0 +1,157 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using FluentAssertions;
+using NUnit.Framework;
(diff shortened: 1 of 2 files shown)


markryd (approved):
I had a look at the history, and pre-netcore we were using Uri.HexEscape, which we then ported, so presumably this was always broken. EscapeDataString and HexEscape give different values for standard chars (HexEscape converts them to hex anyway) but I guess if it works with Octopus Server then we weren't relying on that ¯\_(ツ)_/¯

PR #195, opened 2017-08-30
Adding 'EnhancedMode' NuGet feed resource property

paulmarsy:
I've inherited from NuGetFeedResource to add the property to verify that it works & allows the 'Use extended API' option to be modified. 
extendedapi

https://github.com/paulmarsy/OctopusNuGetDeploymentFeed/blob/3aa99d370529a6a5b533a8ce771a6d192aaeb0aa/OctopusNuGetDeploymentFeed/Octopus/EnhancedNuGetFeedResource.cs


markryd:
Thanks for the PR @paulmarsy

Issue #120, opened 2017-01-28
Output verbosity of logger


patriksvensson:
It would be nice if we could configure the Serilog log level as an argument to octo.exe.

Right now, the pack command (when adding a large directory) writes a lot of debug messages for `Adding file` which I'm not interested in.

If this is something that you think is worth having, I would glady contribute a pull request.

droyad:
![](https://github.trello.services/images/mini-trello-icon.png) [Output verbosity of logger #120](https://trello.com/c/MmscOrwK/208-output-verbosity-of-logger-120)

patriksvensson:
@droyad That wasn't meant as an answer to me right? 😆

droyad:
@patriksvensson That's a "we'll get back to you".  Which I'll do right now.

Sounds like a good idea. It has a few wrinkles, in that the logger is created before options are parsed. Also there is (currently) no place to add that option.

I suggest adding a base class that all the commands inherit from (see `ApiCommand`, which most but not all already inherit from) and adding the option there. Then call `ConfigureLogger` again passing the log level.

patriksvensson:
@droyad I noticed that problem when I looked at the source code yesterday. Your suggestion seems like a good one so I'll try to submit a PR within a couple of days.

Thanks for the quick reply!

PR #201, opened 2017-10-31
Added log level option to commands.

patriksvensson:
This is a proposal for fixing #120.

I'm not very familiar with your code base, so I apologize if it there are discrepancies in code style or similar. I've tried to keep the code as idiomatic to the one in the code base.

If you have any feedback for changes and/or improvements I will happily change my code. : smile:

## ApiCommand based commands

image

## Pack command

image

### DIFF
--- a/source/Octo/Diagnostics/LogUtilities.cs
+++ b/source/Octo/Diagnostics/LogUtilities.cs
@@ -1,0 +1,48 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using Octopus.Cli.Commands;
+using Octopus.Cli.Infrastructure;
+using Serilog.Core;
+using Serilog.Events;
+
+namespace Octopus.Cli.Diagnostics
+{
+    internal static class LogUtilities
+    {
+        private static readonly Dictionary<string, LogEventLevel> Lookup;
+        public static LoggingLevelSwitch LevelSwitch { get; }
+        public static LogEventLevel DefaultLogLevel => LogEventLevel.Debug;
+
+        static LogUtilities()
+        {
+            LevelSwitch = new LoggingLevelSwitch(DefaultLogLevel);
+            Lookup = ((LogEventLevel[]) Enum.GetValues(typeof(LogEventLevel)))
+                .ToDictionary(key => key.ToString().ToLowerInvariant(), value => value);
+        }
+        
+        public static LogEventLevel ParseLogLevel(string value)
+        {
+            if (Lookup.TryGetValue(value, out var level))
+            {
+                return level;
+            }
+
+            throw new CommandException($"Unrecognized loglevel '{value}'. Valid options are {GetValidOptions()}. " + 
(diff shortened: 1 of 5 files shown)


mcasperson:
Hi Patrik, thanks for this PR. We like the idea, but unfortunately in its current for the new logging options don't work as expected.

In Program.cs, this code:

```
                if (command is ILogLevelAware logLevelAwareCommand)
                {
                    if (logLevelAwareCommand.LogLevel != null)
                    {
                        ConfigureLogger(logLevelAwareCommand.LogLevel);
                    }
                }
```

is executed before this code:

```
command.Execute(args.Skip(1).ToArray()).GetAwaiter().GetResult();
```

`command.Execute` is where the command line params are parsed. So prior to `command.Execute`, `logLevelAwareCommand.LogLevel` is always null. This means the log level is never changed.

patriksvensson:
@mcasperson Ok. That’s strange. I will take a look at it but I tried it out and it seemed to work ad expected...

I will take a look and update the PR in a couple hours.

patriksvensson:
@mcaspwrson Ok, I see it now. Seems to be an issue of ”10 unit tests, 0 integration tests” :smile:

The fix should be easy. I’ll update te PR in a couple of hours.

patriksvensson:
@mcasperson Ok, I've updated the PR. It's not very testable in the current state, but I knowingly opted for a simple solution which didn't require us to inject Serilog's `LoggingLevelSwitch` in every command and we can take the discussion from there.

Another approach we could take is to create a decorator for `ILogger` that contains the logging level switch directly. This would make the log level code more testable, but also increase the complexity + maintainability of the code base.

It's possible I've missed something obvious here, if that's the case, please don't hesitate to correct me :smile:

source/Octo/Diagnostics/LogUtilities.cs
```diff
@@ -0,0 +1,54 @@
...
+            LevelSwitch = new LoggingLevelSwitch(LogEventLevel.Debug);
+        }
+        
+        public static LogEventLevel ParseLogLevel(string value)
+        {
+            if (_lookup.TryGetValue(value, out var level))
+            {
+                return level;
+            }
+            throw new CommandException($"Unrecognized loglevel '{value}'.");
```
mcasperson: We try to provide valid options when displaying an error. So this code would print:

> Unrecognized loglevel 'blah'.

Ideally it should print something like:

> Unrecognized loglevel 'blah'. Valid options are fatal, error, warning, information, debug or verbose.

patriksvensson:
@mcasperson I've updated the message per your request and I also did some minor changes to the description of the command to match other commands better.

I've also rebased my branch against master.

mcasperson:
Thanks for that Patrik. We've merged this, and the changes will go out with the next release.

patriksvensson:
@mcasperson Perfect! Do you have any estimation of when you're going to push the new client?

PR #211, opened 2017-12-11
Fix Octo.exe project import problem if channel has version rule

mayuanyang:
Fixes OctopusDeploy/Issues#4037
## Issue
If a project channel has a version rule, it will fail on import when using `Import` command from `Octo.exe`

## Why
Channel version rule requires a `Package Step`, however project importer imports the `Channels` before the `DeploymentProcess`, when Octopus Server try to save the Channel, it looks for the referenced `Package Step` which isn't imported yet and throw exception

## Resolution
Change the import order to save `DeploymentProcess` before `Channels`, then remap the new channel ids back to the newly imported actions

## Tests
- Import a project with channel version rule which is not in the target server
Expect output: Project, Channels and Deployment process imported successfully
- Import a project with channel version rule which exists in the target server
Expect output: Project, Channels and Deployment process imported successfully with the new settings

### DIFF
--- a/source/Octo/Importers/ProjectImporter.cs
+++ b/source/Octo/Importers/ProjectImporter.cs
@@ -1,9 +1,7 @@
 ﻿using System;
 using System.Collections.Generic;
-using System.Diagnostics;
 using System.Globalization;
 using System.Linq;
-using System.Reflection;
 using System.Threading.Tasks;
 using Serilog;
 using Octopus.Cli.Commands;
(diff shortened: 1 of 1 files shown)


michaelnoonan:
[Fix Octo.exe project import failed if channel has version rule](https://trello.com/c/sGPlStUT/870-fix-octoexe-project-import-failed-if-channel-has-version-rule)

PR #216, opened 2018-01-16
Reverted Uri encode function back to original but still handling doub…

benPearce1:
…le byte characters


benPearce1:
Fixes https://github.com/OctopusDeploy/Issues/issues/4170

PR #221, opened 2018-01-24
Added version command (plus test)

jaymickey:
Basic `octo.exe version` command. Simply returns the version number.

### DIFF
--- a/source/Octo.Tests/Commands/VersionTestFixture.cs
+++ b/source/Octo.Tests/Commands/VersionTestFixture.cs
@@ -1,0 +1,55 @@
+using System;
+using System.IO;
+using FluentAssertions;
+using NUnit.Framework;
+using Octopus.Cli.Commands;
+using Octopus.Cli.Tests.Helpers;
+using Octopus.Cli.Util;
+using Serilog;
+
+namespace Octopus.Cli.Tests.Commands
+{
+    [TestFixture]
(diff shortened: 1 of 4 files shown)


source/Octo.Tests/Commands/VersionTestFixture.cs
```diff
@@ -0,0 +1,55 @@
...
+
+            commandOutputProvider = new CommandOutputProvider(logger);
+            versionCommand = new VersionCommand(commandOutputProvider);
+            logger = new LoggerConfiguration().WriteTo.TextWriter(output).CreateLogger();
+        }
+
+        [Test]
+        public void ShouldPrintCorrectVersionNumber()
+        {
+            var version = GetVersionFromFile(Path.Combine(TestContext.CurrentContext.WorkDirectory, "ExpectedSdkVersion.txt"));
```
mcasperson: ExpectedSdkVersion.txt needs to be added to .gitignore as it is not removed after a test, but should not be included in the repo. If it is not excluded and IDE will incorrectly automatically include it at some point.

mcasperson:
This PR is good to go save for ignoring the test text file that gets created.

jaymickey:
Hi @mcasperson,

Absolutely right, I've added change to .gitignore.

mcasperson:
Thanks for that @jaymickey .

PR #224, opened 2018-02-02
Octopus Security Exception helptext

benPearce1:
Will write out the helptext of a security exception in Octo.exe if available.


benPearce1:
Not required

PR #226, opened 2018-02-09
Adding migration models and repositories

thebigmoosey:
We want to start exposing the migration API for hosted, so it's time to start exposing this in OctopusClients.

I also sorted the properties in the main repositories, as this has been bugging my OCD :)


thebigmoosey:
[Adding migration models and repositories to Octopus.Clients](https://trello.com/c/1U8wjgJh/134-adding-migration-models-and-repositories-to-octopusclients)

PR #230, opened 2018-02-16
Add workers to clients lib (not including octo.exe)

MichaelJCompton:
Adds worker machines and pools to the clients lib.  Brings clients up to date with `enh-workers` branch in octopus.

Doesn't address Octo.exe or RegisterMachineOperation.  The PR just gets too big with all that.  So those have been split into other tasks.

https://trello.com/c/MGqEZSlW/2025-workers-octoexe

and

https://trello.com/c/XAb4ye84/2028-workers-tentacle-register-with

**Note:** This is full of breaking changes because it moves type names to match what's now in server.  Is it too much to do this (minus the last commit has the same effects without the name changes)?  Or should it go further and change all the functions that "get machines" etc, to refer to deployment targets.


MichaelJCompton:
closed in favour of https://github.com/OctopusDeploy/OctopusClients/pull/242

PR #239, opened 2018-03-28
Handle SSL Offloading for Auth Cookies

michaelnoonan:
Relates to https://github.com/OctopusDeploy/OctopusDeploy/pull/2029

### DIFF
--- a/source/Octopus.Client/Model/LoginCommand.cs
+++ b/source/Octopus.Client/Model/LoginCommand.cs
@@ -24,5 +24,10 @@ namespace Octopus.Client.Model
         /// Whether the cookie should be persistent.
         /// </summary>
         public bool RememberMe { get; set; }
+
+        /// <summary>
+        /// Encapsulated state for post successful login.
+        /// </summary>
+        public LoginState State { get; set; }
     }
 }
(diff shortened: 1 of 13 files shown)


michaelnoonan:
I wonder if we should fill this in by default? It feels like something the underlying HttpClient should know about... 🤔

source/Octopus.Client/Model/LoginCommand.cs
```diff
@@ -24,5 +24,10 @@ public class LoginCommand
         /// Whether the cookie should be persistent.
         /// </summary>
         public bool RememberMe { get; set; }
+
+        /// <summary>
+        /// Encapsulated state for post successful login.
```
michaelnoonan: I think we should mention this will be set automatically by the client when the request is made, but can be set explicitly if so desired.

source/Octopus.Client/Model/LoginState.cs
```diff
@@ -0,0 +1,12 @@
+﻿namespace Octopus.Client.Model
+{
+    public class LoginState
+    {
+        /// <summary>
+        /// Whether the client says it's using a secure connection. We need this because SSL offloading
+        /// can obscure this and the server cannot tell whether the client initiated the call using a secure
+        /// connection.
+        /// </summary>
+        public bool UsingSecureConnection { get; set; }
```
michaelnoonan: Same here, this will be set automatically by the client, unless it is set explicitly.

PR #247, opened 2018-04-20
Add workers to Octo

MichaelJCompton:
Patterned off similar endpoints for Environments and Machines.

### DIFF
--- a/source/Octopus.Cli/Commands/WorkerPool/CleanWorkerPoolCommand.cs
+++ b/source/Octopus.Cli/Commands/WorkerPool/CleanWorkerPoolCommand.cs
@@ -1,0 +1,176 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+using Octopus.Cli.Infrastructure;
+using Octopus.Cli.Repositories;
+using Octopus.Cli.Util;
+using Octopus.Client;
+using Octopus.Client.Model;
+using Octopus.Client.Model.Endpoints;
+
+namespace Octopus.Cli.Commands.WorkerPools
+{
+    [Command("clean-workerpool", Description = "Cleans all Offline Worker Machines from a WorkerPool")]
+    public class CleanWorkerPoolCommand : ApiCommand, ISupportFormattedOutput
+    {
+        string poolName;
+        readonly HashSet<string> statuses = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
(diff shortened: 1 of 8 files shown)


source/Octopus.Cli/Commands/WorkerPool/CleanWorkerPoolCommand.cs
```diff
@@ -0,0 +1,176 @@
...
+            options.Add("calamari-outdated=", "[Optional] State of Calamari to clean up. By default ignores Calamari state.", v => SetFlagState(v, ref isCalamariOutdated));
+            options.Add("tentacle-outdated=", "[Optional] State of Tentacle version to clean up. By default ignores Tentacle state", v => SetFlagState(v, ref isTentacleOutdated));
+        }
+
+        public async Task Request()
+        {
+            if (string.IsNullOrWhiteSpace(poolName))
+                throw new CommandException("Please specify a worker pool name using the parameter: --workerpool=XYZ");
+            if (!healthStatuses.Any() && !statuses.Any())
+                throw new CommandException("Please specify a status using the parameter: --health-status");
```
alfhenrik: Should this mention the `status` parameter as you can use either or both the `health-status` and `status` parameters to filter the machines.

source/Octopus.Cli/Commands/WorkerPool/CleanWorkerPoolCommand.cs
```diff
@@ -0,0 +1,176 @@
...
+
+            workerpoolResource = await GetWorkerPool().ConfigureAwait(false);
```
alfhenrik: Such nitpick but OCD 😇 should be `workerPoolResource`

source/Octopus.Cli/Commands/WorkerPool/CleanWorkerPoolCommand.cs
MichaelJCompton: I interpreted this https://github.com/OctopusDeploy/OctopusClients/blob/master/source/Octopus.Cli/Commands/HealthStatusProvider.cs#L13-L16 as meaning that the `status` was for backward compatibility.

I thought cause worker pools is new, it doesn't need the backward compatibility ??

MichaelJCompton: ooohh, whoops, if I meant what I said I would have removed the status option.

MichaelJCompton: ok, I'm just confused, will redo.

PR #251, opened 2018-04-23
Find environments and tenants by name

tothegills:
Looping over `FindByName` is much quicker than `FindByNames` when there are a lot of them. `FindByNames` loops over every page of results vs the single request per environment/tenant for `FindByName`.

Relates to OctopusDeploy/Issues#4468


tothegills:
[1000000% performance increase when registering Tentacle](https://trello.com/c/82rEPv2l/70-1000000-performance-increase-when-registering-tentacle)

PR #258, opened 2018-05-17
aligning the resources against server

slewis74:
Removing incorrect divergences that the new test in Octopus server identified.

`AmazonWebServicesRoleAccount` related things removed as they're a special case that is a server only concern. There is only 1, and you can't see/edit them through the UI/API.


zentron:
Nice one. I ended up having to add the `AmazonWebServicesRoleAccount` value in the `Octopus.Server` copy of this enum when removing the `Octopus.Clients` references since something was referencing it. If it's server only im assuming that is fine to leave there? https://github.com/OctopusDeploy/OctopusDeploy/blob/master/source/Octopus.Core/Resources/Accounts/AccountType.cs

thebigmoosey:
Very nice work. We had a branch for the Azure changes, could have sworn they got merged, but there were so many merge issues along the way (on the long-running feature branch) I'm guessing things got overwritten and not noticed in the end

+1 to computers detecting our human insufficiencies :D

slewis74:
@zentron yeah, it's ok to leave in server. That is an account type in the database, but it represents the machine that's running the script so doesn't make sense to let users see/edit it. It's hidden in the UI IIRC.

@michaelnoonan you had a question? About the cert chain structure I'd guess? I mirrored server's structure, and is why I added @MJRichardson as a reviewer cause maybe it's not right :)

PR #260, opened 2018-05-18
Kubernetes Target

mcasperson:


source/Octo/Properties/launchSettings.json
```diff
@@ -1,7 +0,0 @@
-{
-  "profiles": {
```
zentron: Should this be .gitignored?

mcasperson: Yeah, I've excluded this now

PR #262, opened 2018-05-21
Space context switch

mayuanyang:


source/Octopus.Client/OctopusAsyncClient.cs
```diff
@@ -162,10 +164,21 @@ private bool IgnoreServerCertificateCallback(HttpRequestMessage message, X509Cer
...
         public async Task<RootResource> RefreshRootDocument()
         {
-            RootDocument = await Get<RootResource>("~/api").ConfigureAwait(false);
+            RootDocument = await Get<RootResource>(rootDocumentUri).ConfigureAwait(false);
             return RootDocument;
         }
 
+        public async Task<IOctopusAsyncClient> ForSpaceContext(string spaceId)
+        {
+            this.clientOptions.SpaceId = spaceId;
```
Flamage82: Don't mutate the existing `clientOption`.  This could have unintended side effects.  For example, if we needed to re-calculate the rootDocumentUri of the current client.

PR #266, opened 2018-05-23
Support space for CommunityActionTemplate

mayuanyang:


source/Octopus.Client/Repositories/Async/CommunityActionTemplateRepository.cs
```diff
@@ -16,19 +17,36 @@ public CommunityActionTemplateRepository(IOctopusAsyncClient client) : base(clie
...
 
-        public Task<ActionTemplateResource> GetInstalledTemplate(CommunityActionTemplateResource resource)
+        public Task<ActionTemplateResource> GetInstalledTemplate(CommunityActionTemplateResource resource, string spaceId = null)
         {
-            return Client.Get<ActionTemplateResource>(resource.Links["InstalledTemplate"]);
+            return Client.Get<ActionTemplateResource>(resource.Links["InstalledTemplate"].AppendSpaceId(spaceId));
+        }
+    }
+
+    static class LinkSpaceExtension
```
Flamage82: This should be in its own file.  Very hard to find otherwise.

source/Octopus.Client/Repositories/UserRepository.cs
```diff
@@ -88,10 +89,10 @@ public UserResource GetCurrent()
             return Client.Get<UserResource>(Client.RootDocument.Link("CurrentUser"));
         }
 
-        public UserPermissionSetResource GetPermissions(UserResource user)
+        public UserPermissionSetResource GetPermissions(UserResource user, string spaceId)
```
Flamage82: Does this need a default of null like all the other actions?

mayuanyang: good catch

PR #287, opened 2018-07-02
Add API to query mix scope document from multiple spaces

mayuanyang:


source/Octopus.Client/Repositories/Async/IMixScopeRepository.cs
```diff
@@ -0,0 +1,11 @@
+using System;
+using System.Collections.Generic;
+using System.Threading.Tasks;
+
+namespace Octopus.Client.Repositories.Async
+{
+    public interface IMixScopeRepository<TResource>
```
TomPeters: Should be "MixedScope". Also, I don't think we want to expose this abstraction to consumers of the client. It is difficult enough to explain the concept of mixed-scope endpoints internally, let alone to our customers.

source/Octopus.Client/Repositories/Async/MixScopeResourceRepository.cs
```diff
@@ -0,0 +1,36 @@
...
+namespace Octopus.Client.Repositories.Async
+{
+    class MixScopeResourceRepository<TMixScopedResource> : BasicRepository<TMixScopedResource> where TMixScopedResource : class, IResource
+    {
+        public MixScopeResourceRepository(IOctopusAsyncClient client, string collectionLinkName)
+            : base(client, collectionLinkName)
+        {
+        }
+
+        public async Task<List<TMixScopedResource>> Search(bool includeGlobal, string[] spaceIds, object parameters = null)
```
TomPeters: There's more than just `Search` which may need to be scoped to certain spaces/includeGlobal. It really applies for any get/delete.

What about an API that's more like this:

`repository.Teams.LimitTo("spaces-1", "spaces-2").FindAll()`

Any mixed scope repository type would have this `LimitTo` method on it.

mayuanyang: very good idea

source/Octopus.Client/Util/ParameterHelper.cs
```diff
@@ -0,0 +1,34 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Reflection;
+
+namespace Octopus.Client.Util
+{
+    public static class ParameterHelper
+    {
+        public static Dictionary<string, object> CombineParameters(object leftHandSide, object rightHandSide)
```
TomPeters: We might want to be a little bit more specific here about precedence. If the user supplies a `includeGlobal` parameter to a request made within a repository that is already scoped to a certain space context, the repositories context should "win" or overwrite whatever the user has supplied.

source/Octopus.Client/Repositories/BasicRepository.cs
```diff
@@ -15,6 +16,7 @@ abstract class BasicRepository<TResource> where TResource : class, IResource
     {
         readonly IOctopusClient client;
         protected readonly string CollectionLinkName;
+        protected object LimitedToSpacesParameters { get; set; }
```
TomPeters: Can we use a more specific type here that `object`?

source/Octopus.Client/Repositories/BasicRepository.cs
```diff
@@ -47,7 +49,8 @@ public void Delete(TResource resource)
 
         public void Paginate(Func<ResourceCollection<TResource>, bool> getNextPage, string path = null, object pathParameters = null)
         {
-            client.Paginate(path ?? client.RootDocument.Link(CollectionLinkName), pathParameters ?? new { }, getNextPage);
+            var parameters = ParameterHelper.CombineParameters(LimitedToSpacesParameters, pathParameters);
+            client.Paginate(path ?? client.RootDocument.Link(CollectionLinkName), parameters, getNextPage);
```
TomPeters: What about all of the other `GET` methods that require these parameters? eg `GetAll`, and any methods in derived repositories?

source/Octopus.Client/Repositories/Async/TaskRepository.cs
```diff
@@ -223,5 +223,13 @@ public Task WaitForCompletion(TaskResource[] tasks, int pollIntervalSeconds = 4,
         /// <param name="pageSize">Number of items per page, setting to less than the total items still retreives all items, but uses multiple requests reducing memory load on the server</param>
         /// <returns></returns>
         public Task<List<TaskResource>> GetAllActive(int pageSize = int.MaxValue) => FindAll(pathParameters: new { active = true, take = pageSize });
+
+        public ITaskRepository LimitTo(bool includeGlobal, params string[] spaceIds)
```
TomPeters: Given the name `LimitTo`, I would expect this to work:
`taskRepository.LimitTo(true, "Spaces-1", "Spaces-2").LimitTo(true, "Spaces-1")`

but not this:
`taskRepository.LimitTo(true, "Spaces-1", "Spaces-2").LimitTo(true, "Spaces-3")`

In the latter case, the space selection is not a subset of the previous repository scope, so I would expect it to throw an exception.

source/Octopus.Client/Repositories/Async/BasicRepository.cs
```diff
@@ -143,9 +151,31 @@ public Task<TResource> Refresh(TResource resource)
...
+                    ["includeGlobal"] = includeGlobal,
+                    ["spaces"] = spaces
+                };
+            }
+
+            void ValidateSpacesParameters(params string[] spaces)
+            {
+                if (LimitedToSpacesParameters == null) return;
+                var previouslyDefinedSpaceIds = LimitedToSpacesParameters["spaces"] as string[];
+                var previouslyDefinedSpaceIdsSet = new HashSet<string>(previouslyDefinedSpaceIds);
```
TomPeters: Don't cast here... just store a more specific type. Just create your own type, and map to the parameter dictionary at request time

source/Octopus.Client/Repositories/Async/BasicRepository.cs
```diff
@@ -143,9 +151,31 @@ public Task<TResource> Refresh(TResource resource)
...
+                if (!previouslyDefinedSpaceIdsSet.IsSupersetOf(spaces))
```
TomPeters: This should also take into account `includeGlobal`. If you were previously excluding global, and you have tried to limit it to something that now includesGlobal, this should be an exception.

source/Octopus.Client/Repositories/Async/BasicRepository.cs
TomPeters: Ah I see why you did it this way - the basic repository is used by non-space scoped repositories as well. Maybe this indicates we need another base class in that type hierarchy?

source/Octopus.Client/Exceptions/InvalidSpacesLimitationParametersException.cs
```diff
@@ -0,0 +1,11 @@
+﻿using System;
+
+namespace Octopus.Client.Exceptions
+{
+    public class InvalidSpacesLimitationParametersException : OctopusException
+    {
+        public InvalidSpacesLimitationParametersException(): base((int)System.Net.HttpStatusCode.Conflict, "The new spaceIds should be a subset of the previously defined spaceIds")
```
TomPeters: `(int)System.Net.HttpStatusCode.Conflict`

This is a little weird since we haven't fired off a request. I don't see any reason why we should "simulate" a request exception here, so let's just not inherited from `OctopusException`.

source/Octopus.Client/Exceptions/InvalidSpacesLimitationParametersException.cs
```diff
@@ -0,0 +1,11 @@
+﻿using System;
+
+namespace Octopus.Client.Exceptions
+{
+    public class InvalidSpacesLimitationParametersException: Exception
+    {
+        public InvalidSpacesLimitationParametersException(): base("The new spaceIds can only be a subset of the previously defined spaceIds and includeGlobal should be the same as the previous value")
```
TomPeters: `includeGlobal should be the same as the previous value` this isn't true... if it was previously `true` then you should be able to change it to `false`

source/Octopus.Client/Repositories/Async/MixedScopeBaseRepository.cs
```diff
@@ -0,0 +1,35 @@
...
+namespace Octopus.Client.Repositories.Async
+{
+    class MixedScopeBaseRepository<TMixedScopeResource>: BasicRepository<TMixedScopeResource> where TMixedScopeResource : class, IResource
+    {
+        public MixedScopeBaseRepository(IOctopusAsyncClient client, string collectionLinkName) : base(client, collectionLinkName)
+        {
+        }
+
+        SpaceQueryParameters SpaceQueryParameters { get; set; }
+        protected void SetupParameters(bool includeGlobal, string[] spaceIds)
```
TomPeters: Can we pass this in as a constructor parameter instead of modifying an already constructed repo?

source/Octopus.Client/Util/ParameterHelper.cs
```diff
@@ -0,0 +1,58 @@
...
+using Octopus.Client.Exceptions;
+using Octopus.Client.Repositories.Async;
+
+namespace Octopus.Client.Util
+{
+    static class ParameterHelper
+    {
+        public static Dictionary<string, object> CombineParameters(Dictionary<string, object> currentAdditionalQueryParameters, object userProvidedParameters)
+        {
+            if (currentAdditionalQueryParameters == null)
```
TomPeters: You can make the default implementation in `BasicRepository` return an empty dictionary to avoid this null check

source/Octopus.Client/Util/ParameterHelper.cs
```diff
@@ -0,0 +1,58 @@
...
+            if (userProvidedParameters == null)
+                userProvidedParameters = new { };
+            var resultDictionary = GetParameters(userProvidedParameters);
+
+            // Value from the current repository(currentAdditionalQueryParameters) overrides the one from user (userProvidedParameters)
+            currentAdditionalQueryParameters.ToList().ForEach(x => resultDictionary[x.Key] = x.Value);
+            return new Dictionary<string, object>(resultDictionary);
+        }
+
+        public static void ValidateSpaceParameters(SpaceQueryParameters curreSpaceQueryParameters, bool newIncludeGlobal,
```
TomPeters: This method should probably take in two instances of `SpaceQueryParameters` and compare them. Feels a little strange that you are essentially passing the same type of data to this method in two different ways.

source/Octopus.Client/Util/ParameterHelper.cs
```diff
@@ -0,0 +1,58 @@
...
+            string[] newSpaceIds)
+        {
+            if (curreSpaceQueryParameters == null) return;
+
+            if (curreSpaceQueryParameters.IncludeGlobal != newIncludeGlobal)
```
TomPeters: This isn't right -> if you are current `includeGlobal = true`, then you should be able to limit this to `includeGlobal = false`

source/Octopus.Client/Util/ParameterHelper.cs
```diff
@@ -0,0 +1,58 @@
...
+            {
+                throw new InvalidSpacesLimitationParametersException();
```
TomPeters: You could actually throw a different exception in these different cases. `IncludeGlobal` changing from `false` to `true` is one exception, providing a non-subset of spaces is a different exception.

source/Octopus.Client/Repositories/Async/EventRepository.cs
```diff
@@ -116,6 +117,15 @@ public EventRepository(IOctopusAsyncClient client)
                 toAutoId,
                 documentTypes
             });
+
+            return Client.List<EventResource>(Client.RootDocument.Link("Events"), parameters);
+        }
+
+        public IEventRepository LimitTo(bool includeGlobal, params string[] spaceIds)
+        {
+            var repository = new EventRepository(Client);
```
TomPeters: It feels like this should be more like
```
var parameters = this.CreateParameters(spaceIds, includeGlobal);
return new EventRepository(Client, parameters); // this can be a private constructor
```

where `CreateParameters` in a protected method that looks like

```
protected SpaceQueryParameters CreateParameters(bool includeGlobal, parameters string[] spaceIds) 
{
    var newParameters = new SpaceQueryParameters(){IncludeGlobal = includeGlobal, SpaceIds = spaceIds};
    ParameterHelper.ValidateSpaceParameters(SpaceQueryParameters, newParameters);
    return newParameters;
}
```

source/Octopus.Client/Repositories/Async/MixedScopeBaseRepository.cs
```diff
@@ -0,0 +1,35 @@
...
+            ParameterHelper.ValidateSpaceParameters(SpaceQueryParameters, includeGlobal, spaceIds);
+            SpaceQueryParameters = new SpaceQueryParameters(){IncludeGlobal = includeGlobal, SpaceIds = spaceIds};
+        }
+
+        protected override Dictionary<string, object> AdditionalQueryParameters
+        {
+            get
+            {
+                if (SpaceQueryParameters == null)
+                    return null;
```
TomPeters: We can return an empty dictionary here and make the contract of `AdditionalQueryParameters` a little more strict in that it can't be null. We control all implementations so this should be safe.

mayuanyang:
Good points, that was meant to be a bit more refactoring but hasn't got a chance yet, will do it later today

TomPeters:
no rush, don't let it distract you from the testathon :)

source/Octopus.Client/Util/ParameterHelper.cs
```diff
@@ -0,0 +1,54 @@
...
+            if (userProvidedParameters == null)
+                userProvidedParameters = new { };
+            var resultDictionary = GetParameters(userProvidedParameters);
+
+            // Value from the current repository(currentAdditionalQueryParameters) overrides the one from user (userProvidedParameters)
+            currentAdditionalQueryParameters.ToList().ForEach(x => resultDictionary[x.Key] = x.Value);
+            return new Dictionary<string, object>(resultDictionary, StringComparer.OrdinalIgnoreCase);
+        }
+
+        public static void ValidateSpaceParameters(SpaceQueryParameters currentSpaceQueryParameters, SpaceQueryParameters newSpaceQueryParameters)
```
TomPeters: This feels like it should be in `MixedScopeBaseRepository`. I get why you have pulled it out though (because we have duplicated code between the sync and async repos). We have a lot of duplicate code actually, so this is not a new problem :(. The thing I don't like about this implementation is that it makes an assumption that `currentSpaceQueryParameters` can be null, which feels like an implementation detail of `MixedScopeBaseRepository`. I think if we can pull the null check up, that would be much better.

source/Octopus.Client/Exceptions/IncludeGlobalCannotBeSetFromFalseToTrueException.cs
```diff
@@ -0,0 +1,11 @@
+﻿using System;
+
+namespace Octopus.Client.Exceptions
+{
+    public class IncludeGlobalCannotBeSetFromFalseToTrueException : Exception
```
TomPeters: This exception doesn't really feel like it helps the user understand why this is an invalid operation. The name also feels a bit odd.

Maybe something like `public InvalidIncludeGlobalConfigurationException(): base("Cannot include global resources when global resources have already been excluded")

source/Octopus.Client/Exceptions/InvalidSpacesLimitationParametersException.cs
```diff
@@ -0,0 +1,11 @@
+﻿using System;
+
+namespace Octopus.Client.Exceptions
+{
+    public class InvalidSpacesLimitationParametersException: Exception
+    {
+        public InvalidSpacesLimitationParametersException(): base("The new spaceIds can only be a subset of the previously defined spaceIds")
```
TomPeters: I think the word `Parameters` here is redundant. Let's just make it `InvalidSpacesLimitationException`

PR #307, opened 2018-09-10
Spaces api refactor

mayuanyang:
## What
- Instead of having `ICanLimitTo` for mixed scope, we now have `ICanSpaceContext`
- Instead of having spaceId as the constructor input, they now take`SpaceContext`. Default space + system is used if not provided
- Mixed scope repository can be extend to support multiple spaces + system
- Task and CommunityActionTemplate repos no longer take spaceId as input, instead it will check if it can get it from the `SpaceContext`, or throw for write request in multi context
- CookieContainer is preserved after space context switch

### DIFF
--- a/source/Octopus.Client/Repositories/Async/BasicRepository.cs
+++ b/source/Octopus.Client/Repositories/Async/BasicRepository.cs
@@ -85,7 +85,7 @@ namespace Octopus.Client.Repositories.Async
 
         public Task<List<TResource>> GetAll()
         {
-            var parameters = ParameterHelper.CombineParameters(AdditionalQueryParameters, new { id = "all" });
+            var parameters = ParameterHelper.CombineParameters(AdditionalQueryParameters, new { id = IdValueConstant.IdAll });
             return Client.Get<List<TResource>>(Client.Link(CollectionLinkName), parameters);
         }
 
--- a/source/Octopus.Client/Repositories/BasicRepository.cs
+++ b/source/Octopus.Client/Repositories/BasicRepository.cs
@@ -88,7 +88,7 @@ namespace Octopus.Client.Repositories
 
         public List<TResource> GetAll()
         {
-            var parameters = ParameterHelper.CombineParameters(AdditionalQueryParameters, new { id = "all" });
+            var parameters = ParameterHelper.CombineParameters(AdditionalQueryParameters, new { id = IdValueConstant.IdAll });
             return client.Get<List<TResource>>(client.Link(CollectionLinkName), parameters);
         }
 
--- a/source/Octopus.Client/MixedScopeConstants.cs
+++ b/source/Octopus.Client/MixedScopeConstants.cs
@@ -6,4 +6,9 @@
         public const string QueryStringParameterIncludeSystem = "includeSystem";
         public const string AllSpacesQueryStringParameterValue = "all";
     }
+
+    internal static class IdValueConstant
+    {
+        public const string IdAll = "all";
+    }
 }
--- a/source/Octopus.Client/Repositories/Async/MachineRepository.cs
+++ b/source/Octopus.Client/Repositories/Async/MachineRepository.cs
@@ -76,7 +76,7 @@ namespace Octopus.Client.Repositories.Async
         public Task<List<MachineResource>> FindByThumbprint(string thumbprint)
         {
             if (thumbprint == null) throw new ArgumentNullException("thumbprint");
-            return Client.Get<List<MachineResource>>(Client.Link("machines"), new { id = "all", thumbprint });
+            return Client.Get<List<MachineResource>>(Client.Link("machines"), new { id = IdValueConstant.IdAll, thumbprint });
         }
 
         /// <summary>
--- a/source/Octopus.Client/Repositories/Async/TenantRepository.cs
+++ b/source/Octopus.Client/Repositories/Async/TenantRepository.cs
@@ -39,7 +39,7 @@ namespace Octopus.Client.Repositories.Async
         /// <returns></returns>
         public Task<List<TenantResource>> FindAll(string name, string[] tags, int pageSize = int.MaxValue)
         {
-            return Client.Get<List<TenantResource>>(Client.Link("Tenants"), new { id = "all", name, tags, take = pageSize });
+            return Client.Get<List<TenantResource>>(Client.Link("Tenants"), new { id = IdValueConstant.IdAll, name, tags, take = pageSize });
         }
 
         public Task<TenantVariableResource> ModifyVariables(TenantResource tenant, TenantVariableResource variables)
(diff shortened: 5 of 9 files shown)


source/Octopus.Client/Repositories/Async/UserRepository.cs
```diff
@@ -68,13 +68,14 @@ public async Task<UserResource> Register(RegisterCommand registerCommand)
...
-        public Task SignIn(LoginCommand loginCommand)
+        public async Task SignIn(LoginCommand loginCommand)
         {
             if (loginCommand.State == null)
             {
                 loginCommand.State = new LoginState { UsingSecureConnection = Client.IsUsingSecureConnection };
             }
-            return Client.Post(Client.Link("SignIn"), loginCommand);
+            await Client.Post(Client.Link("SignIn"), loginCommand);
+            await Client.ReloadRootDocumentsAfterUserSignedIn();
```
mayuanyang: There is a use case where a client initialized as unauthenticated and then sign in later. As the client now will load space and user info which require authentication and permissions, spaceRootDocument is not populated for unauthenticated client and get reloaded after user signed in.

source/Octopus.Client/OctopusAsyncClient.cs
```diff
@@ -36,11 +37,14 @@ public class OctopusAsyncClient : IOctopusAsyncClient
         bool ignoreSslErrorMessageLogged = false;
         private readonly string rootDocumentUri;
         private OctopusClientOptions clientOptions;
+        private bool signedIn = false;
+        private bool IsAuthenticated => (signedIn || !string.IsNullOrEmpty(this.serverEndpoint.ApiKey));
```
NickJosevski: 👍

source/Octopus.Client/OctopusAsyncClient.cs
```diff
@@ -686,6 +724,50 @@ private async Task<T> ReadResponse<T>(HttpResponseMessage response)
...
+        }
+
+        private void ValidateSpaceId(string spaceId)
+        {
+            if (string.IsNullOrEmpty(spaceId))
+            {
+                throw new ArgumentException("spaceId cannot be null");
+            }
+
+            if (spaceId == "all")
```
NickJosevski: hmmm, maybe at least put `"all"` in a global const string somewhere, cause it's also used on line 755?

source/Octopus.Client/OctopusClient.cs
```diff
@@ -648,14 +701,42 @@ protected virtual OctopusResponse<TResponseResource> DispatchRequest<TResponseRe
...
             }
         }
 
+        private void ValidateSpaceId(string spaceId)
+        {
+            if (string.IsNullOrEmpty(spaceId))
+            {
+                throw new ArgumentException("spaceId cannot be null");
+            }
+            if (spaceId == "all")
```
NickJosevski: same as above, put this definition somewhere similar to how we have `DefaultIdentifiers` in Octopus server

source/Octopus.Client/Repositories/Async/BasicRepository.cs
```diff
@@ -11,150 +12,203 @@
...
-                {
-                    resources.AddRange(page.Items.Where(search));
-                    return true;
-                }, path, pathParameters)
-                    .ConfigureAwait(false);
-                return resources;
-            }
+        public Task<List<TResource>> GetAll()
+        {
+            var parameters = ParameterHelper.CombineParameters(AdditionalQueryParameters, new { id = "all" });
```
NickJosevski: can use the `all` const string here too

source/Octopus.Client/Repositories/Async/BasicRepository.cs
```diff
@@ -11,150 +12,203 @@
...
-                    })
-                    .ConfigureAwait(false);
-
-                return resources;
-            }
+        bool IsIncludingSystem()
+        {
+            var isMixedScope = TypeUtil.IsAssignableToGenericType(this.GetType(), typeof(ICanExtendSpaceContext<>));
+            return isMixedScope
+                ? bool.Parse(AdditionalQueryParameters["includeSystem"].ToString())
```
NickJosevski: while you're cleaning up some strings throughout if you make a static class with some defaults, can you put the `"includeSystem"` string in there too, and find and replace the usages too, a typo in one of these would be a pain to track down and fix in the future.

source/Octopus.Client/Repositories/Async/BasicRepository.cs
```diff
@@ -11,150 +12,203 @@
...
 
-            public Task<TResource> Refresh(TResource resource)
+        void ValidateSpaceId(TResource resource)
+        {
+            if (resource is IHaveSpaceResource spaceResource)
             {
-                if (resource == null) throw new ArgumentNullException("resource");
-                return Get(resource.Id);
+                var spaceIds = GetSpaceIds();
+                var isWildcard = spaceIds != null && spaceIds.Contains("all");
```
NickJosevski: another case of `"all"`

can you name this differently, we have a concept of wildcard in permissions, just so there's no way to confuse the concepts.

source/Octopus.Client/Repositories/BasicRepository.cs
```diff
@@ -156,6 +159,55 @@ public TResource Refresh(TResource resource)
...
+                ? bool.Parse(AdditionalQueryParameters["includeSystem"].ToString())
+                : Client.SpaceContext.IncludeSystem;
+        }
+
+        void ValidateSpaceId(TResource resource)
+        {
+            if (resource is IHaveSpaceResource spaceResource)
+            {
+                var spaceIds = GetSpaceIds();
+                var isWildcard = spaceIds != null && spaceIds.Contains("all");
```
NickJosevski: another case of `"all"`

can you name this differently, we have a concept of wildcard in permissions, just so there's no way to confuse the concepts.

source/Octopus.Client/SpaceContext.cs
```diff
@@ -0,0 +1,39 @@
...
+    public class SpaceContext
+    {
+        public static SpaceContext SpecificSpace(string spaceId) => new SpaceContext(new [] {spaceId}, false);
+        public static SpaceContext SpecificSpaceAndSystem(string spaceId) => new SpaceContext(new []{spaceId}, true);
+        public static SpaceContext SystemOnly() => new SpaceContext(new string[0], true);
+
+        internal SpaceContext(IReadOnlyCollection<string> spaceIds, bool includeSystem)
+        {
+            if (spaceIds.Count == 0 && !includeSystem)
+                throw new ArgumentException("At least 1 spaceId is required when includeSystem is set to false");
```
NickJosevski: should this also throw `MismatchSpaceContextException`?

mayuanyang: I feel `ArgumentException` is a bit more appropriate as it does not mismatch with anything yet, it is some invalid arguments being passed in that cause the `SapceContext` become not usable, what do you think? @NickJosevski

NickJosevski: I see yea that's fine then.

PR #310, opened 2018-09-13
User permissions in Spaces

pawelpabich:
Client side changes for user permissions


pawelpabich:
[User permissions API should be multi-scope](https://trello.com/c/4r2xkZDI/229-user-permissions-api-should-be-multi-scope)

PR #318, opened 2018-09-26
Push space specific implementation from client to repository

mayuanyang:
This PR has break changes
- We make OctopusRepository's constructor as private and use `Create` to initialize, that way we can do async, the reason behind is we want to load spaces the user can access during initialization.
- To make it a bit more consistent, we now keep `SpaceRootDocument` and `RootDocument` in the repository and deprecated the `RootDocument` from client.


source/Octopus.Client/IOctopusClient.cs
```diff
@@ -13,6 +13,8 @@ namespace Octopus.Client
     /// </summary>
     public interface IOctopusClient : IDisposable
     {
+        IOctopusRepository Repository { get; }
```
mayuanyang: Current `master` does not provide a pre-initialized `Repository` for handy, we add one here. This lines up with `IOctopusAsyncClient`

source/Octopus.Client/OctopusAsyncRepository.cs
```diff
@@ -134,5 +156,138 @@ public OctopusAsyncRepository(IOctopusAsyncClient client)
         public IWorkerRepository Workers { get; }
         public IScopedUserRoleRepository ScopedUserRoles { get; }
         public IUserPermissionsRepository UserPermissions { get; }
+
+        public async Task<IOctopusAsyncRepository> ForSpaceContext(string spaceId)
+        {
+            ValidateSpaceId(spaceId);
+            var spaceRoot = await LoadSpaceRootResource(Client, RootDocument, spaceId);
+            return new OctopusAsyncRepository(Client, RootDocument, spaceRoot, SpaceContext.SpecificSpace(spaceId));
```
mayuanyang: `RootDocument` is shared

source/Octopus.Client/IOctopusAsyncClient.cs
```diff
@@ -396,5 +354,7 @@ public interface IOctopusAsyncClient : IDisposable
         /// </summary>
         /// <returns></returns>
         Task SignOut();
+
+        bool IsAuthenticated { get; }
```
TomPeters: Why do we need this new public method where we didn't need it before?

source/Octopus.Client/IOctopusAsyncRepository.cs
```diff
@@ -9,60 +11,87 @@ namespace Octopus.Client
...
+        IServerStatusRepository ServerStatus { get; }
+        ISpaceRepository Spaces { get; }
+        IUserRepository Users { get; }
+        IUserRolesRepository UserRoles { get; }
+        /// <summary>
+        /// The client over which the repository is run.
+        /// </summary>
+        IOctopusAsyncClient Client { get; }
+        SpaceContext SpaceContext { get; }
+        Task<IOctopusAsyncRepository> ForSpaceContext(string spaceId);
```
TomPeters: BTW this method should just be named `ForSpace`... remove the word `Context` here

source/Octopus.Client/OctopusAsyncClient.cs
```diff
@@ -150,110 +144,39 @@ private static async Task<IOctopusAsyncClient> Create(OctopusServerEndpoint serv
...
-            {
-                RootResource = rootResource;
-                SpaceRootResource = spaceRootResource;
-            }
-
-            public RootResource RootResource { get; }
-            public SpaceRootResource SpaceRootResource { get; }
-        }
+        [Obsolete("This property is deprecated, please use the one from Repository instead")]
+        public RootResource RootDocument => Repository?.RootDocument;
```
TomPeters: I think this delegation should be the other way around - we store the root document on the client (deprecated), and the repository delegates to the root document on the client.

This way, if people are `new`ing up the client and repository manually (rather than doing `client.Repository`), then they can still access the root document on either the client or the repository. For those people that are `new`ing it up manually in existing code, they won't be doing `client.Repository = repository`

source/Octopus.Client/OctopusAsyncRepository.cs
TomPeters: It would be better to pass the root repository around instead of the root document. That way, if you refresh the root repository, this change will propagate to all linked child repositories.

source/Octopus.Client/OctopusAsyncClient.cs
```diff
@@ -150,110 +144,39 @@ private static async Task<IOctopusAsyncClient> Create(OctopusServerEndpoint serv
...
         {
-            var rootDocument = await Get<RootResource>(rootDocumentUri).ConfigureAwait(false);
-            var spaceRootDocument = await LoadSpaceRootResource(this, rootDocument);
-            RootDocuments = new RootResources(rootDocument, spaceRootDocument);
-            return rootDocument;
+            return await Repository.RefreshRootDocument();
         }
 
-        public async Task<IOctopusAsyncClient> ForSpace(string spaceId)
+        protected virtual Task<RootResource> EstablishSession()
```
TomPeters: Do you know why this guy is protected virtual? Looks pretty weird and this is a kind of breaking change.

TomPeters:
Now is probably a good time to get #team-integration involved before we do any more work on this

source/Octopus.Client/IOctopusAsyncClient.cs
mayuanyang: Previously the client only gets the RootDocument (which allows anonymous) from the server during initialization. We are now also getting user and the spaces he/she can access to which requires authentication,  https://github.com/OctopusDeploy/OctopusClients/pull/318/files#diff-1b64c49f341db495d9a8b5a38fc1903cR281

source/Octopus.Client/OctopusAsyncClient.cs
mayuanyang: Not too sure, I cannot find the usage of it, but I might not know enough

TomPeters:
This has been superseded by a number of other PRs. Spaces support is already fully implemented.

PR #319, opened 2018-09-28
Make the client work with ALL THE SERVERS

pawelpabich:
I would not call it a supported feature but it might give some of our biggest clients time to migrate all of their Octopus Servers to Spaces without having to use different version of the client with different versions of the server.


pawelpabich:
[C# client crashes script in Permanent Instance](https://trello.com/c/3iIqLUhm/345-c-client-crashes-script-in-permanent-instance)

PR #325, opened 2018-10-11
Spaces client

mayuanyang:
## High level of interfaces architecture for the spaces
This is just the idea, DO NOT MERGE

## Summary
- Separate repositories for System vs space contexts, exposed through new `ForSystem()` and `ForSpace()` methods on the client.
- A combined repository for backwards compatibility and for users relying on the default space (i.e. you can still go `client.Repository`). Note that we have made the decision to always have a default space enabled for new installations, but still give users the option of disabling it.
- Mixed scope repositories (events, tasks, teams) can be extended to include a wider collection of spaces, which reflects that you can query these endpoints across spaces (like `/api/events?spaces=Spaces-1,Spaces-2,Spaces-2&includeSystem=true`)
- `RootDocument` moved to the System Repository, since it is an system-wide API concern
- `RootDocument` still exists on the client, but is marked as `Obsolete`. Implementation will delegate to the repository.
- Space repository will expose the `SpaceRootDocument` (not shown in this PR)
- Removed `RefreshRootDocument()` (is this ok?). It feels like we would need to add support for the analogous `RefreshSpaceRootDocument()` as well, which makes things a little uglier. It feels like this method should be required really rarely, and the user can always just create a new client/repository to get around this problem.

### DIFF
--- a/source/Octopus.Client/IAsyncSpaceRepository.cs
+++ b/source/Octopus.Client/IAsyncSpaceRepository.cs
@@ -1,0 +1,41 @@
+using Octopus.Client.Repositories.Async;
+
+namespace Octopus.Client
+{
+    public interface IAsyncSpaceRepository
+    {
+        IAccountRepository Accounts { get; }
+        IActionTemplateRepository ActionTemplates { get; }
+        IArtifactRepository Artifacts { get; }
+        IBuiltInPackageRepositoryRepository BuiltInPackageRepository { get; }
+        ICertificateConfigurationRepository CertificateConfiguration { get; }
+        ICertificateRepository Certificates { get; }
+        IChannelRepository Channels { get; }
+        IDashboardConfigurationRepository DashboardConfigurations { get; }
+        IDashboardRepository Dashboards { get; }
+        IDefectsRepository Defects { get; }
+        IDeploymentProcessRepository DeploymentProcesses { get; }
+        IDeploymentRepository Deployments { get; }
+        IEnvironmentRepository Environments { get; }
+        IFeedRepository Feeds { get; }
+        IInterruptionRepository Interruptions { get; }
+        ILibraryVariableSetRepository LibraryVariableSets { get; }
+        ILifecyclesRepository Lifecycles { get; }
+        IMachinePolicyRepository MachinePolicies { get; }
+        IMachineRepository Machines { get; }
+        IMachineRoleRepository MachineRoles { get; }
+        IProjectGroupRepository ProjectGroups { get; }
+        IProjectRepository Projects { get; }
+        IProjectTriggerRepository ProjectTriggers { get; }
+        IProxyRepository Proxies { get; }
+        IReleaseRepository Releases { get; }
+        IRetentionPolicyRepository RetentionPolicies { get; }
+        ISubscriptionRepository Subscriptions { get; }
+        ITagSetRepository TagSets { get; }
+        ITenantRepository Tenants { get; }
+        ITenantVariablesRepository TenantVariables { get; }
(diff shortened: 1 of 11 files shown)


source/Octopus.Client/IOctopusAsyncRepository.cs
```diff
@@ -7,59 +10,37 @@ namespace Octopus.Client
...
+        /// HTTP 401 or 403: Thrown when the current user's API key was not valid, their
+        /// account is disabled, or they don't have permission to perform the specified action.
+        /// </exception>
+        /// <exception cref="OctopusServerException">
+        /// If any other error is successfully returned from the server (e.g., a 500
+        /// server error).
+        /// </exception>
+        /// <exception cref="OctopusValidationException">HTTP 400: If there was a problem with the request provided by the user.</exception>
+        /// <exception cref="OctopusResourceNotFoundException">HTTP 404: If the specified resource does not exist on the server.</exception>
+        RootResource RootDocument { get; }
```
TomPeters: I would put this on `IAsyncSystemRepository`. Consider customers that have disabled the default space: they should ideally never see or use this particular repository (`IOctopusAsyncRepository `) and should instead use the system/space variants instead.

source/Octopus.Client/IOctopusAsyncRepository.cs
```diff
@@ -7,59 +10,37 @@ namespace Octopus.Client
...
+        /// <summary>
+        /// Requests a fresh root document from the Octopus Server which can be useful if the API surface has changed. This can occur when enabling/disabling features, or changing license.
+        /// </summary>
+        /// <returns>A fresh copy of the root document.</returns>
+        Task<RootResource> RefreshRootDocument();
```
TomPeters: Do you think we could remove this entirely?

This also makes sense to be added to the Space repository and the space root document since it could change for similar reasons (i.e. `Task<SpacerootResource> RefreshSpaceRootDocument();`). This feels a little ugly.

There is a really easy workaround: create a new instance of the client/repository.

source/Octopus.Client/IAsyncSpaceRepository.cs
```diff
@@ -0,0 +1,41 @@
+using Octopus.Client.Repositories.Async;
+
+namespace Octopus.Client
+{
+    public interface IAsyncSpaceRepository
```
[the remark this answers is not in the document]
TomPeters: Mixed scope concerns (tasks/events/teams) should probably be here too.

source/Octopus.Client/IAsyncSystemRepository.cs
```diff
@@ -0,0 +1,19 @@
+using Octopus.Client.Repositories.Async;
+
+namespace Octopus.Client
+{
+    public interface IAsyncSystemRepository
```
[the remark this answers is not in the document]
TomPeters: Mixed scope concerns (tasks/events/teams) should probably be here too.

source/Octopus.Client/IOctopusAsyncRepository.cs
```diff
@@ -7,59 +9,11 @@ namespace Octopus.Client
     /// Functionality not exposed by this interface can be accessed
     /// using <see cref="IOctopusAsyncRepository.Client" />.
     /// </summary>
-    public interface IOctopusAsyncRepository
+    public interface IOctopusAsyncRepository : IAsyncSpaceRepository, IAsyncSystemRepository
```
jburger: Very minor thing - `IAsyncSpaceRepository` seems inconsistently named from `IOctopusAsyncRepository` - wondering if `ISpaceAsyncRepository` is more consistent?

PR #328, opened 2018-10-16
Add support for "octopack" file

BlythMeister:
An octopack contains various command line arguments and means you can
store the structure of your octopus package in a repo rather than as a
set of command line arguments
The format of this is similar to a paket.template file

Update existing pack tests to include tests for octopack files


BlythMeister:
Closing as this actually doesn't do what i wanted. :(

PR #329, opened 2018-10-16
Add ability to tell octo push to ignore errors about package existing

BlythMeister:
If we get a 409 error (CONFLICT) and we are doing "IgnoreExisting"
Then get all packages with that name & check for the version number matching.
If matching, do not error.
If no matching, throw.


thedewi:
Sorry this didn't get followed up here, @BlythMeister. I want to let you know that support has been added in Octo.exe 6.9.0, and youll get this behaviour if you use: `--overwrite-mode=IgnoreIfExists`

The rest of the documentation is here: https://octopus.com/docs/octopus-rest-api/octo.exe-command-line/push

PR #334, opened 2018-10-23
Space client

mayuanyang:
## What
We want to maximise the backward compatibility and also make users fall into the pit of success.

## Changes
- Introduce `IOctopusSpaceAsyncRepository` and `IOctopusSystemAsyncRepository`
- Every specific repository now depends on `IOctopusRepository/IOctopusAsyncRepository` rather than `IOctopusClient/IOctopusAsyncClient`
- Add `ForSpace` and `ForSystem` to `IOctopusClient/IOctopusAsyncClient`
- Remove `RootDocument` from client and lazy load it from Repository


source/Octopus.Client/OctopusAsyncRepository.cs
```diff
@@ -134,5 +142,118 @@ public OctopusAsyncRepository(IOctopusAsyncClient client)
         public IWorkerRepository Workers { get; }
         public IScopedUserRoleRepository ScopedUserRoles { get; }
         public IUserPermissionsRepository UserPermissions { get; }
+
+        private SpaceRootResource SpaceRootDocument { get; set; }
+        private RootResource RootDocument { get; set; }
+
+        public async Task<bool> HasLink(string name)
+        {
+            await LoadRootDocuments();
```
TomPeters: Anywhere where you've introduced an `await` within this project, you should also use `.ConfigureAwait(false)`

source/Octopus.Client/OctopusAsyncClient.cs
```diff
@@ -135,126 +140,37 @@ public static async Task<IOctopusAsyncClient> Create(OctopusServerEndpoint serve
...
-            {
-                RootResource = rootResource;
-                SpaceRootResource = spaceRootResource;
-            }
-
-            public RootResource RootResource { get; }
-            public SpaceRootResource SpaceRootResource { get; }
+            // User used to see this exception 
+            // System.PlatformNotSupportedException: The handler does not support custom handling of certificates with this combination of libcurl (7.29.0) and its SSL backend
+            await client.Repository.LoadRootDocument();
```
TomPeters: Ah this is annoying. Is there a better way to do this? Or can we remove this switch altogether now?

source/Octopus.Client/OctopusAsyncClient.cs
```diff
@@ -27,25 +26,19 @@ namespace Octopus.Client
...
-        private OctopusClientOptions clientOptions;
-        private bool signedIn = false;
-        private bool IsAuthenticated => (signedIn || !string.IsNullOrEmpty(this.serverEndpoint.ApiKey));
-
-        // Use the Create method to instantiate
-        private OctopusAsyncClient(OctopusServerEndpoint serverEndpoint, OctopusClientOptions options, bool addCertificateCallback)
+        private bool ignoreSslErrorMessageLogged = false;
+        private string antiforgeryCookieName = null;
+
+        public OctopusAsyncClient(OctopusServerEndpoint serverEndpoint, OctopusClientOptions options, bool addCertificateCallback)
```
TomPeters: Given that `Create` is still required to do async work ( 😢 ) this should probably still be private...

source/Octopus.Client/OctopusAsyncClient.cs
```diff
@@ -135,126 +140,37 @@ public static async Task<IOctopusAsyncClient> Create(OctopusServerEndpoint serve
...
-            var spaceRoot = await LoadSpaceRootResource(this, RootDocument);
-            RootDocuments = new RootResources(RootDocument, spaceRoot);
+            await Post(await Repository.Link("SignIn"), loginCommand);
+
+            // Capture the cookie name here so that the Dispatch method does not rely on the rootDocument to get the InstallationId
+            antiforgeryCookieName = cookieContainer.GetCookies(cookieOriginUri)
+                .Cast<Cookie>()
+                .Single(c => c.Name.StartsWith(ApiConstants.AntiforgeryTokenCookiePrefix)).Name;
+
+            Repository = new OctopusAsyncRepository(this);
```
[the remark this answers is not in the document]
mayuanyang: Might be a extreme case, if the current logged in user is has default space [spaces-1] on, follow by the default space is switched off/or a different space is now become the default [spaces-2], when calling the `SignIn` method again, `SpaceRootDocument` might need to reload

PR #338, opened 2018-10-30
Spaces octo exe

mayuanyang:
## What
It will now take a optional `--space` which is the space name as parameter and setup the `Repository` for that space, resources will then be created, modified and executed in that space context


source/Octopus.Cli/Commands/ApiCommand.cs
```diff
@@ -68,6 +69,7 @@ protected ApiCommand(IOctopusClientFactory clientFactory, IOctopusAsyncRepositor
             options.Add("proxy=", $"[Optional] The URI of the proxy to use, eg http://example.com:8080.", v => clientOptions.Proxy = v);
             options.Add("proxyUser=", $"[Optional] The username for the proxy.", v => clientOptions.ProxyUsername = v);
             options.Add("proxyPass=", $"[Optional] The password for the proxy. If both the username and password are omitted and proxyAddress is specified, the default credentials are used. ", v => clientOptions.ProxyPassword = v);
+            options.Add("space=", $"[Optional] The name of a space, resource will be created, executed within the context of the given space, default space will be used if it is omitted. ", v => spaceName = v);
```
TomPeters:
```suggestion
            options.Add("space=", $"[Optional] The name of a space within which this command will be executed. The default space will be used if it is omitted. ", v => spaceName = v);
```

source/Octopus.Cli/Commands/ApiCommand.cs
```diff
@@ -138,7 +140,23 @@ public async Task Execute(string[] commandLineArguments)
...
+                if (space == null)
+                    throw new CommandException($"Cannot find the space with name {spaceName}");
+
+                Repository = repositoryFactory.CreateRepository(client, RepositoryScope.ForSpace(space.Id));
+                commandOutputProvider.Debug("Space name specified, process is now running in the context of space: {space:l}", spaceName);
+            }
+            else
+            {
+                Repository = repositoryFactory.CreateRepository(client);
+                commandOutputProvider.Debug("Space name unspecified, process will try to run in the default space context if it is enabled");
```
TomPeters:
```suggestion
                commandOutputProvider.Debug("Space name unspecified, process will try to run in the default space context if it is enabled, or in backwards compatible mode for older versions of Octopus Server");
```

source/Octopus.Cli/Commands/ApiCommand.cs
mayuanyang: 👍

source/Octopus.Cli/Commands/ApiCommand.cs
mayuanyang: 👍

PR #342, opened 2018-11-01
Spaces tweak repository scope

TomPeters:
- Need to be able to get the global octopus server certificates without stepping into a space scope
- You can't currently actually install community action templates into a non-default space, because the community action template repository is only exposed on the system repository.

### DIFF
--- a/source/Octopus.Client/IOctopusCommonAsyncRepository.cs
+++ b/source/Octopus.Client/IOctopusCommonAsyncRepository.cs
@@ -9,5 +9,6 @@ namespace Octopus.Client
         ITeamsRepository Teams { get; }
         IScopedUserRoleRepository ScopedUserRoles { get; }
         IUserPermissionsRepository UserPermissions { get; }
+        ICommunityActionTemplateRepository CommunityActionTemplates { get; }
     }
 }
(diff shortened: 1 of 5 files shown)


source/Octopus.Client/IOctopusCommonAsyncRepository.cs
```diff
@@ -9,5 +9,6 @@ public interface IOctopusCommonAsyncRepository
         ITeamsRepository Teams { get; }
         IScopedUserRoleRepository ScopedUserRoles { get; }
         IUserPermissionsRepository UserPermissions { get; }
+        ICommunityActionTemplateRepository CommunityActionTemplates { get; }
```
TomPeters: This means you can go `client.ForSystem().Install(...)` and it will always fail at runtime.

To make this fail at compile time involves a breaking change, so it needs further thought. It might still be worth doing, in which case we can do this as a separate PR.

TomPeters: To be clear, the breaking change would involve splitting up the community action template repository into two repositories. See https://github.com/OctopusDeploy/OctopusClients/commit/225b8e302069d67f5ecc320e19e05403a407f4dd

mayuanyang: Good catch

PR #346, opened 2018-11-05
Renaming DeploymentTarget things

slewis74:
Use DeploymentTarget instead of Machine where we really mean the target. This is a step towards less confusion when we physcially separate Machines and Deployment Targets


slewis74:
Closing as we're rethinking this

PR #349, opened 2018-11-15
EnsureSystemContext for certain tasks

johnsimons:


source/Octopus.Client/Repositories/MixedScopeBaseRepository.cs
```diff
@@ -109,5 +107,30 @@ protected void EnsureSingleSpaceContext()
...
+        protected void EnsureSystemContext()
+        {
+            var exception = new Exception("Attempted to perform a system operation in a space scoped context. Ensure you are in system context first by using client.ForSystem()");
+            Repository.Scope.Apply(_ => throw exception,
+                () => { },
+                () => 
+                {
+                    if (userDefinedSpaceContext == null)
+                    {
+                        return; // Assumes the default space
```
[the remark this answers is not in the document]
johnsimons: fixed

source/Octopus.Client/Repositories/MixedScopeBaseRepository.cs
```diff
@@ -109,5 +107,30 @@ protected void EnsureSingleSpaceContext()
...
+                    }
+
+                    userDefinedSpaceContext.ApplySpaceSelection(spaces =>
```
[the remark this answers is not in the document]
johnsimons: fixed

johnsimons:
@TomPeters regarding https://github.com/OctopusDeploy/OctopusClients/pull/349#pullrequestreview-175185047, it is a tricky test to do,
we try to do things for the user, but it changes the behaviour of the code too much IMO.
Example:
```
var taskRepository = repository.Tasks.UsingContext(SpaceContext.SpecificSpaces("Spaces-2"));

var spaceTask = taskRepository.Create(new TaskResource {SpaceId = "Spaces-1", Name = BuiltInTasks.Health.Name, Description = string.Empty});
spaceTask.SpaceId.Should().Be("Spaces-1");
```
This actually fails, because of https://github.com/OctopusDeploy/OctopusClients/blob/6a7d0e9c5b06e68c72f239ab3cabddc067de40d4/source/Octopus.Client/Repositories/MixedScopeBaseRepository.cs#L66-L69

TomPeters:
The problem is that Tasks suck in general because we have different data (different tasks) but we are pretending the data is homogenous, and this introduces a whole lot of edge cases in the tasks repository. I think it is right for the complexity to be isolated there.

In the example you gave, I would either expect it to fail (because the SpaceId you provided is different from the context you are in) or for it to just work, by ignoring the spaceId you provided on the resource (which is the current implementation). We decided to go with the latter option, which could be stated as:

`If we are able to automatically populate the SpaceId property, then we should ignore whatever the user has provided and populate it for them`.

This means the scenario you have tested should succeed and create a task in `Spaces-2` and this is expected behaviour.

The alternative is to require the SpaceId property on the resource to match the context. This sucks in the following case:

```
var taskRepository = repository.Tasks.UsingContext(SpaceContext.SpecificSpaces("Spaces-2"));

var spaceTask = taskRepository.Create(new TaskResource {Name = BuiltInTasks.Health.Name}); // fails
```

This fails because the task has a null SpaceId (the user hasn't specified it). However, a null `SpaceId` indicates that the task should be a system task, which is invalid based on the current space context. Automatically setting the spaceId felt like a better behaviour in this scenario.

So I'm proposing that we throw an exception if we *can't* automatically set the SpaceId. This happens when you are in some multi-scope context and the SpaceId property on the resource does not match your current context.

Yes, the complexity here sucks but we have to live with some complexity unless we want to break backwards compatibility.

source/Octopus.Client/Repositories/Async/MixedScopeBaseRepository.cs
```diff
@@ -104,5 +102,21 @@ protected void EnsureSingleSpaceContext()
                     }, () => throw new SingleSpaceOperationInMultiSpaceContextException());
                 });
         }
+
+        protected void EnsureSystemContext()
+        {
+            var exception = new Exception("Attempted to perform a system operation in a space scoped context. Ensure you are in system context first by using client.ForSystem()");
```
TomPeters: Can we create a more specific exception type here? Better to use specific types for libraries

johnsimons:
> In the example you gave, I would either expect it to fail (because the SpaceId you provided is different from the context you are in) or for it to just work, by ignoring the spaceId you provided on the resource (which is the current implementation). We decided to go with the latter option,

I am not sure this is a safe default IMO, it could mean they end-up modifying a different space from what they intended, I think it is safer to throw and let the user know that they have made a mistake, I don't think we can make a safe assumption in this situation

source/Octopus.Client/Repositories/Async/MixedScopeBaseRepository.cs
TomPeters: Yeah I guess you're right, people are unlikely to programatically handle any custom exception here. Can we at least use a more specific built-in exception type? eg `InvalidOperationException`?

johnsimons: fixed

johnsimons:
I'll merge this and update https://github.com/OctopusDeploy/OctopusDeploy/pull/3077 to use it

PR #353, opened 2018-11-29
Add AutomationEnvironment to the User-Agent

slewis74:
Update to OctopusClient so it includes the BuildEnvironment in the User-Agent header when it recognises it's being used by a TeamCity, Bamboo or Azure DevOps build agent.

The header is now formatted as follows:

`OctopusClient-dotnet/version (system-information) AutomationEnvironment[/version][ octo[ plugin/version]]`

where:
system-information conveys info about the OS environment
AutomationEnvironment is something like NoneOrUnknown, TeamCity, Bamboo, AzureDevOps, or Octopus
`octo` is added if OctopusClient is being used from the `octo.exe` command line.
'plugin' is added if octo is being called from one of our build server extensions.


source/Octopus.Client.Tests/Integration/OctopusClient/BuildServerHeadersAzureDevOpsTests.cs
```diff
@@ -0,0 +1,11 @@
+using Octopus.Client.Model;
+
+namespace Octopus.Client.Tests.Integration.OctopusClient
+{
+    public class BuildServerHeadersAzureDevOpsTests : BuildServerHeadersBaseTests
+    {
+        protected override string EnvironmentVariableName => OctopusCustomHeaders.EnvVar_AzureDevOps;
```
matt-richardson: Potentially coupled too closely to the implementation here - ie, the value could change and the test would pass, but it'd break when parsing the telemetry.

source/Octopus.Client.Tests/Integration/OctopusClient/BuildServerHeadersBaseTests.cs
```diff
@@ -0,0 +1,61 @@
...
+            });
+        }
+
+        protected abstract string EnvironmentVariableName { get; }
+        protected abstract string EnvironmentVariableValue { get; }
+        internal abstract BuildEnvironment ExpectedBuildEnvironment { get; }
+
+        protected override void SetupEnvironmentVariables()
+        {
+            OctopusCustomHeaders.GetEnvironmentVariable = variableName => variableName == EnvironmentVariableName ? EnvironmentVariableValue : null;
```
matt-richardson: this is a little bit of a brain bender

source/Octopus.Client.Tests/Integration/OctopusClient/UserAgentTests.cs
```diff
@@ -22,11 +23,22 @@ public UserAgentTests()
...
+            OctopusCustomHeaders.GetEnvironmentVariable = Environment.GetEnvironmentVariable;
+        }
+
+
         [Test]
         public async Task AsyncClient_ShouldProvideUserAgent_WithNameAndVersion()
         {
             var response = await AsyncClient.Get<TestDto>(TestRootPath);
-            response.UserAgentValue.Should().Be($"{ApiConstants.OctopusUserAgentProductName}/{GetType().GetSemanticVersion().ToNormalizedString()}", "We should set the standard User-Agent header");
+            response.UserAgentValue.Should().Be($"{ApiConstants.OctopusUserAgentProductName}/{GetType().GetSemanticVersion().ToNormalizedString()} {BuildEnvironment.Unspecified}", "We should set the standard User-Agent header");
```
matt-richardson: Again, I'd posit that this is too tightly coupled...

source/Octopus.Client/Model/BuildEnvironment.cs
```diff
@@ -0,0 +1,10 @@
+﻿namespace Octopus.Client.Model
+{
+    internal enum BuildEnvironment
+    {
+        Unspecified,
```
matt-richardson: Worth spending half an hour to hack in other options?
Maybe use https://github.com/pvdlg/env-ci as inspiration?

source/Octopus.Client/OctopusAsyncClient.cs
```diff
@@ -126,9 +125,27 @@ public static async Task<IOctopusAsyncClient> Create(OctopusServerEndpoint serve
 #endif
         }
 
-        private static async Task<IOctopusAsyncClient> Create(OctopusServerEndpoint serverEndpoint, OctopusClientOptions options, bool addHandler)
+        internal static async Task<IOctopusAsyncClient> Create(OctopusServerEndpoint serverEndpoint, OctopusClientOptions options, string requestingTool)
         {
-            var client = new OctopusAsyncClient(serverEndpoint, options ?? new OctopusClientOptions(), addHandler);
+#if HTTP_CLIENT_SUPPORTS_SSL_OPTIONS
```
matt-richardson: This looks unrelated to the PR?

source/Octopus.Client/OctopusClientFactory.cs
```diff
@@ -8,6 +8,13 @@ namespace Octopus.Client
     /// </summary>
     public class OctopusClientFactory : IOctopusClientFactory
     {
+        private static string requestingTool;
```
matt-richardson: Not enthused by static methods like this. It'd be better to pass it in in `CliProgram` when configuring the builder.

source/Octopus.Client/Model/OctopusCustomHeaders.cs
```diff
@@ -0,0 +1,49 @@
+﻿using System;
+using Octopus.Client.Extensions;
+
+namespace Octopus.Client.Model
+{
+    internal class OctopusCustomHeaders
+    {
+        internal static readonly string EnvVar_TeamCity = "TEAMCITY_VERSION";
```
matt-richardson: Consider a hashtable / dictionary here, so you can just look through that in `DetermineBuildEvironment` instead of an if for every case

source/Octopus.Client/Model/BuildEnvironment.cs
matt-richardson: I'd definitely add in Jenkins, appveyor, bitbucket, travis at least. The more the merrier...

source/Octopus.Client.Tests/Integration/OctopusClient/BuildServerHeadersBaseTests.cs
slewis74: Yes. Yes it is. I had to do this, because it relies on Environment variables that our build servers will always set to be the ones for TeamCity, because that's what they are :) So the live code uses the real `Environment.GetEnvironmentVariable` as that function delegate and the tests can override to mock the environment variables

source/Octopus.Client.Tests/Integration/OctopusClient/BuildServerHeadersAzureDevOpsTests.cs
slewis74: Yep, good point, I'll change that

source/Octopus.Client/Model/BuildEnvironment.cs
slewis74: yep, I'll add the extras.

source/Octopus.Client/OctopusAsyncClient.cs
slewis74: it is related, I added a new private method above the original and also added a new defaulted parameter to the original. Git is visualising that in a confusing manner

source/Octopus.Client/Model/OctopusCustomHeaders.cs
slewis74: 👍

PR #369, opened 2019-01-09
Handling when we are running against an old Server

johnsimons:
Better error messages when default space is off

### DIFF
--- a/source/Octopus.Cli/Commands/ApiCommand.cs
+++ b/source/Octopus.Cli/Commands/ApiCommand.cs
@@ -146,7 +146,7 @@ namespace Octopus.Cli.Commands
             {
                 if (!serverHasSpaces)
                 {
-                    throw new CommandException($"The server {endpoint.OctopusServer} has no spaces. Invoke Octo tool without specifying the space name as an argument");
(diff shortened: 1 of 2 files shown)


source/Octopus.Cli/Commands/ApiCommand.cs
```diff
@@ -140,21 +140,45 @@ public async Task Execute(string[] commandLineArguments)
             commandOutputProvider.PrintHeader();
 
             var client = await clientFactory.CreateAsyncClient(endpoint, clientOptions).ConfigureAwait(false);
+            var serverHasSpaces = await client.ForSystem().HasLink("Spaces").ConfigureAwait(false);
 
             if (!string.IsNullOrEmpty(spaceName))
             {
+                if (!serverHasSpaces)
+                {
+                    throw new CommandException($"The server {endpoint.OctopusServer} has no spaces. Invoke Octo tool without specifying the space name as an argument");
```
jburger: Minor suggestion,
```suggestion
                    throw new CommandException($"The server {endpoint.OctopusServer} has no spaces. Try invoking the Octo tool without specifying the space name as an argument");
```

PR #384, opened 2019-02-11
Workaround for location header issue

BlythMeister:


willsmith9182:
https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/preprocessor-directives/preprocessor-if - consider using an if directive to only target effected framework versions?

PR #385, opened 2019-02-14
Syntax property is valid only for inline scripts

pawelpabich:


pawelpabich:
[Consolidate file shares into one](https://trello.com/c/GA71IuyM/205-consolidate-file-shares-into-one)

PR #393, opened 2019-02-22
New param for import destination package feed

NickJosevski:


NickJosevski:
the octo exe export, doesn't support the full fledged partial export/import process as what's in the full API, so it doesn't need this argument.

PR #395, opened 2019-02-26
Create release and deploy validation

mcasperson:
Fixes https://github.com/OctopusDeploy/Issues/issues/5329

### DIFF
--- a/source/Octopus.Cli/Commands/Deployment/DeploymentCommandBase.cs
+++ b/source/Octopus.Cli/Commands/Deployment/DeploymentCommandBase.cs
@@ -119,7 +119,22 @@ namespace Octopus.Cli.Commands.Deployment
                     throw new CommandException(
                         $"Could not find the tenant {tenantName} on the Octopus Server");
                 }
-            }   
+            } 
+            
+            // Make sure environment is valid
+            foreach (var environmentName in DeployToEnvironmentNames)
+            {
+                var environments = await Repository.Environments.FindByNames(DeployToEnvironmentNames).ConfigureAwait(false);
+                var missingEnvironment = DeployToEnvironmentNames
+                    .Where(env => environments.All(env2 => env2.Name != env))
+                    .ToList();
+                if (missingEnvironment.Count != 0)
+                {
+                    throw new CommandException(
+                        $"The environment{(missingEnvironment.Count == 1 ? "" : "s")} {string.Join(", ", missingEnvironment)} " +
+                        $"do{(missingEnvironment.Count == 1 ? "es" : "")} not exist or {(missingEnvironment.Count == 1 ? "is" : "are")} misspelled");
+                }
+            }
             
             // Make sure the machines are valid
             await GetSpecificMachines();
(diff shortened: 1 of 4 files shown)


source/Octopus.Cli/Commands/Deployment/DeploymentCommandBase.cs
```diff
@@ -17,6 +17,7 @@ namespace Octopus.Cli.Commands.Deployment
 {
     public abstract class DeploymentCommandBase : ApiCommand
     {
+        private const char Seperator = '/';
```
johnsimons: Spelling mistake, `Separator`

source/Octopus.Cli/Commands/Deployment/DeploymentCommandBase.cs
```diff
@@ -81,11 +82,68 @@ protected override async Task ValidateParameters()
...
+            {
+                // Verify the format of the tag
+                var parts = tenantTag.Split(Seperator);
+                if (parts.Length != 2 || string.IsNullOrEmpty(parts[0]) || string.IsNullOrEmpty(parts[1]))
+                {
+                    throw new CommandException(
+                        $"Canonical Tag Name expected in the format of `TagSetName{Seperator}TagName`");
+                }
+                // Verify the presence of the tag
+                var tagSets = await Repository.TagSets.FindByName(parts[0]).ConfigureAwait(false);
```
johnsimons: This should be done outside the loop otherwise we can be retrieving the same tag more than once

source/Octopus.Cli/Commands/Deployment/DeploymentCommandBase.cs
```diff
@@ -81,11 +82,68 @@ protected override async Task ValidateParameters()
...
+                        $"Could not find the tenant {tenantName} on the Octopus Server");
+                }
+            } 
+            
+            // Make sure environment is valid
+            foreach (var environmentName in DeployToEnvironmentNames)
+            {
+                var environments = await Repository.Environments.FindByNames(DeployToEnvironmentNames).ConfigureAwait(false);
+                var missingEnvironment = DeployToEnvironmentNames
+                    .Where(env => environments.All(env2 => env2.Name != env))
```
johnsimons: I am not sure about the impl of `All(predicate)`, but maybe it makes sense to call `All` outside the loop?

mcasperson: Actually the outer loop was not required at all here, so it will be removed.

source/Octopus.Cli/Commands/Deployment/DeploymentCommandBase.cs
mcasperson: Doh! Good catch.

source/Octopus.Cli/Commands/Deployment/DeploymentCommandBase.cs
johnsimons: And this one @mcasperson ?

mcasperson: I've just added a change to cache old queries of the same tagset.

PR #400, opened 2019-03-22
Fix package metadata links

TomPeters:
Fixes https://trello.com/c/0xULFGKG/56-pushing-metadata-ignores-your-space-selection


source/Octopus.Client/Repositories/Async/PackageMetadataRepository.cs
```diff
@@ -15,8 +15,8 @@ public PackageMetadataRepository(IOctopusAsyncRepository repository)
 
         public async Task<OctopusPackageMetadataMappedResource> Get(string id)
         {
-            var rootDocument = await repository.Client.Repository.LoadRootDocument();
```
TomPeters: The root document's links will contain fully qualified links for the default space, for backwards compatibility. If the default space is disabled, then the root document won't even contain these links.

PR #407, opened 2019-04-05
Add Octo.exe argument --noDeployAfter to set QueueTimeExpiry

thedewi:
This adds `--noDeployAfter` to `Octo.exe` which sets `QueueTimeExpiry` on the API, similar to `--deployAt` which sets `QueueTime`.

The original issue suggested the name `--dontStartAfter` but I thought `--noDeployAfter` sounded a little more idiomatic and consistent with the other option. Let me know if a better name springs to mind.

Fixes OctopusDeploy/Issues#5159.


thedewi:
Note: This change impacts docs, so we'll want to get them deployed too.

PR #412, opened 2019-04-15
Added Sever Health Check API

pawelpabich:


pawelpabich:
@michaelnoonan, this API doesn't require auth but the API that provides the link to it does.  I'm think we should expose the link to `Health` endpoint in the `HomeResponder`, what do you think ?

image

michaelnoonan:
I think adding a deep link directly from the home resource makes sense :+1:

pawelpabich:
👍

PR #421, opened 2019-05-06
Ensure resource space scope matches the repository scope

jburger:
This PR attempts to ensure that the user has good advice if they try and modify a resource in a different scope to the repository  they are working with.

[Internal tracking](https://trello.com/c/qEBhmQm7/236-add-validation-to-octopusclient-when-the-spaceid-from-the-route-does-not-match-the-one-from-the-resource-and-throw-exception-if)


source/Octopus.Client/Exceptions/ResourceSpaceDoesNotMatchRepositorySpaceException.cs
```diff
@@ -0,0 +1,18 @@
...
+using Octopus.Client.Model;
+
+namespace Octopus.Client.Exceptions
+{
+    public class ResourceSpaceDoesNotMatchRepositorySpaceException : Exception
+    {
+        public ResourceSpaceDoesNotMatchRepositorySpaceException(
+            IHaveSpaceResource spaceResource, 
+            SpaceResource repositorySpace) 
+            : base($"The resource has a different space specified than the one specified by the repository scope. " +
```
TomPeters:
```suggestion
            : base($"The resource has a different space specified than the one specified by the repository. " +
```

source/Octopus.Client/Exceptions/SpaceResourceIsIncompatibleWithSystemRepositoryException.cs
```diff
@@ -0,0 +1,18 @@
+using System;
+using Octopus.Client.Extensibility;
+
+namespace Octopus.Client.Exceptions
+{
+    public class SpaceResourceIsIncompatibleWithSystemRepositoryException : Exception
```
TomPeters: I believe this is now unused

source/Octopus.Client/Exceptions/SystemResourceIsIncompatibleWithSpaceScopedRepositoryException.cs
```diff
@@ -0,0 +1,18 @@
+using System;
+using Octopus.Client.Extensibility;
+
+namespace Octopus.Client.Exceptions
+{
+    public class SystemResourceIsIncompatibleWithSpaceScopedRepositoryException : Exception
```
TomPeters: I believe this is now unused

source/Octopus.Client.Tests/Repositories/Async/BasicRepositoryFixture.cs
```diff
@@ -0,0 +1,257 @@
...
+using NSubstitute;
+using NUnit.Framework;
+using Octopus.Client.Exceptions;
+using Octopus.Client.Extensibility;
+using Octopus.Client.Model;
+
+namespace Octopus.Client.Tests.Repositories.Async
+{
+    [TestFixture]
+    public class BasicRepositoryFixture
```
TomPeters: The more I look at these tests, the more I think it would be better to do all the testing within `Octopus.Server`. We really just want to assert that the user gets a nice error message - whether this happens within Octopus Client or Octopus Server is an implementation detail that our tests should be agnostic to. Ideally we should be able to remove all of these tests.

jburger: I would like them in one place too, however I preferred them here during development because the feedback loop was tighter. If anything, I submit that we remove them in Octopus Server code.

In question form, what benefit do we gain from having these tests in integration tests, if we are only asserting that the user gets nice error messages?

Additionally, the server side tests do not use both 'flavours' of repository (it only exercises the synchronous client) I feel that having these tests here, reminds us that there are two places to consider, at all times.

jburger: > reminds us that there are two places to consider
...and if that smells - maybe we fix that :)

TomPeters: Jim and I discussed this over zoom. We decided that
- It is reasonable to keep these tests here for the quick feedback loop
- The tests in Octopus Server make the most sense because they test the actual expected behaviour for end users.
- Jim will ask a wider audience about why we are maintaining two separate class hierarchies for the async vs sync repositories.

jburger: We had a quick discussion to resolve this, and we both agree that there are benefits in having tests in both places for now, so these will stay here for now. We may look at addressing the duplicate code issue in a seperate PR down the line.

jburger:
@TomPeters thanks for the feedback! Ready for a final review.

PR #423, opened 2019-05-10
At long last added support for getting and setting the License

droyad:


droyad:
[Octopus.Client - Licenses API](https://trello.com/c/kmQSNaYe/338-octopusclient-licenses-api)

source/Octopus.Client/Repositories/LicensesRepository.cs
```diff
@@ -0,0 +1,31 @@
...
+    {
+        public LicensesRepository(IOctopusRepository repository)
+            : base(repository, "CurrentLicense")
+        {
+        }
+
+        public LicenseResource GetCurrent()
+            => Client.Get<LicenseResource>(Repository.Link(CollectionLinkName));
+
+        public LicenseResource UpdateCurrent(LicenseResource resource)
```
pawelpabich: `BasicRepository` has `Modify` method which does what `UpdateCurrent` does here.  Should we use instead ?

source/Octopus.Client/Model/LicenseStatusResource.cs
```diff
@@ -0,0 +1,64 @@
...
+{
+    public static class LicenseKinds
+    {
+        public static readonly string Unlicensed = "Unlicensed";
+        public static readonly string Trial = "Trial";
+        public static readonly string CommunityEdition = "CommunityEdition";
+        public static readonly string CommercialLicense = "CommercialLicense";
+    }
+    
+    public static class LicenseMessageDispositions
```
pawelpabich: This is an enum on the server side. Are you using `String`s here instead of the enum to avoid problems in the future when the enum gets extended ?

source/Octopus.Client/Repositories/LicensesRepository.cs
droyad: I did consider that, however I decided to go this direction so that it compliments `GetCurrent`. Also if we ever add updating arbitrary licences (probably never), the `Update` method would still be available.

source/Octopus.Client/Model/LicenseStatusResource.cs
droyad: That is right

PR #424, opened 2019-05-16
new permissions to bring proj v-sets in line with new behavior of lib…

NickJosevski:
… v-sets


NickJosevski:
This approach was superceded, by just having scoping on existing ones.

PR #429, opened 2019-06-11
Support deserializing package publish date using multiple formats

zentron:
This change ensures that the `PackageResource` model matches what the Octopus server is now returning for the `PublishDate` which is a nullable DateTimeOffset.  Related server change https://github.com/OctopusDeploy/OctopusDeploy/pull/3937

For older clients reading the server, there is no real breaking change since they are just reading string. For new clients reading old servers,  the multi-deserialization logic provided in this PR will convert both new server or old server responses to a DateTimeOffset property.

### DIFF
--- a/source/Octopus.Client.Tests/Serialization/MultiIsoDateTimeFormatConverterFixture.cs
+++ b/source/Octopus.Client.Tests/Serialization/MultiIsoDateTimeFormatConverterFixture.cs
@@ -1,0 +1,49 @@
+using System;
+using Newtonsoft.Json;
+using NUnit.Framework;
+using Octopus.Client.Serialization;
+
+namespace Octopus.Client.Tests.Serialization
+{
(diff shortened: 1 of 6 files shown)


source/Octopus.Client.Tests/Serialization/MultiIsoDateTimeFormatConverterFixture.cs
```diff
@@ -0,0 +1,49 @@
+using System;
+using Newtonsoft.Json;
+using NUnit.Framework;
+using Octopus.Client.Serialization;
+
+namespace Octopus.Client.Tests.Serialization
+{
+    [TestFixture]
```
mcasperson: A link back to https://github.com/OctopusDeploy/Issues/issues/5535 would be handy here to document why we have two dates.

PR #437, opened 2019-07-05
Overwrite mode

slewis74:
Adding OverwriteMode, to replace the `replace` boolean in the push APIs. Adds the ability for the caller to IgnoreExisting.

Relates to OctopusDeploy/Octopus-TeamCity#18

### DIFF
--- a/source/Octopus.Client/Repositories/Async/BuiltInPackageRepositoryRepository.cs
+++ b/source/Octopus.Client/Repositories/Async/BuiltInPackageRepositoryRepository.cs
@@ -133,7 +133,7 @@ namespace Octopus.Client.Repositories.Async
                     object pathParameters;
                     
                     // if the link doesn't contain overwritemode then we're connected to an older server, which uses the `replace` parameter  
-                    if (link.Contains("overwritemode"))
+                    if (link.Contains(OverwriteModeLink.Link))
                     {
                         pathParameters = new {replace = overwriteMode == OverwriteMode.OverwriteExisting, packageId, signatureResult.BaseVersion};
                     }
--- a/source/Octopus.Client/Repositories/Async/BuiltInPackageRepositoryRepository.cs
+++ b/source/Octopus.Client/Repositories/Async/BuiltInPackageRepositoryRepository.cs
@@ -74,10 +74,10 @@ namespace Octopus.Client.Repositories.Async
             var link = await repository.Link("PackageUpload").ConfigureAwait(false);
             object pathParameters;
 
-            // if the link doesn't contain overwritemode then we're connected to an older server, which uses the `replace` parameter  
+            // if the link contains overwriteMode then we're connected to a new server, if not use the old `replace` parameter  
             if (link.Contains(OverwriteModeLink.Link))
             {
-                pathParameters = new { overwritemode = overwriteMode };
+                pathParameters = new { overwriteMode = overwriteMode };
             }
             else
             {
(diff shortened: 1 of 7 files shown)


source/Octopus.Client/Repositories/BuiltInPackageRepositoryRepository.cs
```diff
@@ -57,18 +69,32 @@ public PackageFromBuiltInFeedResource PushPackage(string fileName, Stream conten
             }
                 
             contents.Seek(0, SeekOrigin.Begin);
+
+            var link = repository.Link("PackageUpload");
+            object pathParameters;
+
+            // if the link doesn't contain overwritemode then we're connected to an older server, which uses the `replace` parameter  
+            if (link.Contains("overwritemode"))
```
hnrkndrssn: Should this be using `OverwriteModeLink.Link`?

source/Octopus.Client/Repositories/Async/BuiltInPackageRepositoryRepository.cs
```diff
@@ -62,18 +71,31 @@ public async Task<PackageFromBuiltInFeedResource> PushPackage(string fileName, S
                 Logger.Info("Pushing the complete package to the server, as delta compression was explicitly disabled");
             }
 
+            var link = await repository.Link("PackageUpload").ConfigureAwait(false);
+            object pathParameters;
+
+            // if the link doesn't contain overwritemode then we're connected to an older server, which uses the `replace` parameter  
+            if (link.Contains("overwritemode"))
+            {
+                pathParameters = new {replace = overwriteMode == OverwriteMode.OverwriteExisting};
```
hnrkndrssn: This logic looks opposite to what I'd expect, i.e. if server support `overwriteMode` we should pass in `overwriteMode` in the path params...

source/Octopus.Client/Repositories/Async/BuiltInPackageRepositoryRepository.cs
```diff
@@ -62,18 +71,31 @@ public async Task<PackageFromBuiltInFeedResource> PushPackage(string fileName, S
                 Logger.Info("Pushing the complete package to the server, as delta compression was explicitly disabled");
             }
 
+            var link = await repository.Link("PackageUpload").ConfigureAwait(false);
+            object pathParameters;
+
+            // if the link doesn't contain overwritemode then we're connected to an older server, which uses the `replace` parameter
```
hnrkndrssn: This comment makes me think you initially wanted to do `if (!link.Contains)` here...?

source/Octopus.Client/Repositories/BuiltInPackageRepositoryRepository.cs
```diff
@@ -57,18 +69,32 @@ public PackageFromBuiltInFeedResource PushPackage(string fileName, Stream conten
...
+            {
+                pathParameters = new {replace = overwriteMode == OverwriteMode.OverwriteExisting};
```
hnrkndrssn: Looks like the logic is reversed here too

source/Octopus.Client/Repositories/Async/BuiltInPackageRepositoryRepository.cs
```diff
@@ -62,18 +71,31 @@ public async Task<PackageFromBuiltInFeedResource> PushPackage(string fileName, S
                 Logger.Info("Pushing the complete package to the server, as delta compression was explicitly disabled");
             }
 
+            var link = await repository.Link("PackageUpload").ConfigureAwait(false);
+            object pathParameters;
+
+            // if the link doesn't contain overwritemode then we're connected to an older server, which uses the `replace` parameter  
+            if (link.Contains("overwritemode"))
```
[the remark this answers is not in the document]
thedewi: Agree, it should be.

source/Octopus.Client/Repositories/Async/BuiltInPackageRepositoryRepository.cs
thedewi: Agree, most of these blocks have the `overwriteMode` case first, and this one should too.

source/Octopus.Client/Repositories/BuiltInPackageRepositoryRepository.cs
thedewi: Agree.

source/Octopus.Client/Repositories/BuiltInPackageRepositoryRepository.cs
thedewi: Agree, most of these have the `overwriteMode` case first, so I recommend swapping this one.

thedewi:
fixed those 4 issues in #439

PR #448, opened 2019-08-06
OverwriteMode warning when using older server

slewis74:
Log a warning if `OverwriteMode.IgnoreIfExists` gets used when connected to an older version of server

Relates to OctopusDeploy/Octopus-TeamCity#18

### DIFF
--- a/source/Octopus.Client/Model/OverwriteMode.cs
+++ b/source/Octopus.Client/Model/OverwriteMode.cs
@@ -1,3 +1,5 @@
+using Octopus.Client.Logging;
+
 namespace Octopus.Client.Model
 {
     public enum OverwriteMode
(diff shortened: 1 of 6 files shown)


source/Octopus.Client/OctopusAsyncRepository.cs
```diff
@@ -4,6 +4,7 @@
 using System.Net.Http;
 using System.Threading.Tasks;
 using Octopus.Client.Exceptions;
+using Octopus.Client.Logging;
```
michaelnoonan: Is this needed?

source/Octopus.Client/Model/OverwriteMode.cs
```diff
@@ -10,5 +12,16 @@ public enum OverwriteMode
     internal static class OverwriteModeLink
     {
         public static string Link = "overwriteMode";
+
+        public static bool AsLegacyReplaceFlag(this OverwriteMode @this, ILog logger)
```
michaelnoonan: Consider a more explicit name like this? I felt more confident of understanding the intent of the code as I imagined a name like this:
```suggestion
        public static bool ConvertToLegacyReplaceFlag(this OverwriteMode @this, ILog logger)
```

source/Octopus.Client/Model/OverwriteMode.cs
```diff
@@ -10,5 +12,16 @@ public enum OverwriteMode
...
+        {
+            if (@this == OverwriteMode.IgnoreIfExists)
+            {
+                logger.Warn("You have selected to ignore existing versions but the Octopus server you are connected to doesn't not support this. Falling back to default value of `replace=false`");
```
michaelnoonan:
```suggestion
                logger.Warn("The option to ignore existing versions is only supported by Octopus Server 2019.XYZ or newer. If you want to use this option please upgrade your Octopus Server. In the meantime we are automatically falling back to the default option `replace=false` for this request.");
```

PR #449, opened 2019-08-08
Push metadata error handling

slewis74:
Throw a more meaningful exception if the server doesn't support pushing metadata.

Relates to OctopusDeploy/Octopus-TeamCity#31


matt-richardson:
@slewis74 - pushed a change to use a specific exception, to make it easier for downstream clients to catch and log. What do you think?

slewis74:
Yeah, looks good

PR #450, opened 2019-08-20
Adding support for cloning a tenant and tenant description

matt-richardson:
Relates to OctopusDeploy/Issues#5727


source/Octopus.Client/Editors/Async/TenantEditor.cs
```diff
@@ -39,6 +39,28 @@ public async Task<TenantEditor> CreateOrModify(string name)
 
             return this;
         }
+        
+        public async Task<TenantEditor> CreateOrModify(string name, string description, string cloneId = null)
```
MJRichardson: I think we should add `param` comments.  As a user I possibly don't know what `cloneId` is.

source/Octopus.Client/Model/TenantResource.cs
```diff
@@ -64,5 +64,9 @@ public TenantResource ClearProjects()
         }
 
         public string SpaceId { get; set; }
+        
+        public string Description { get; set; }
+        
+        public string ClonedFromTenantId { get; set; }
```
MJRichardson: Should `CloneFromTenantId` be read-only?

source/Octopus.Client/Editors/Async/TenantEditor.cs
matt-richardson: Even though xml comments are evil... I concur 👍

source/Octopus.Client/Model/TenantResource.cs
matt-richardson: How does one do that?

thedewi:
One suggestion: we could check whether the Server version supports cloning, and warn the user if they try it somewhere unsupported.

It could work similarly to the approach used recently for overwrite mode, here: https://github.com/OctopusDeploy/OctopusClients/blob/master/source/Octopus.Client/Repositories/Async/BuiltInPackageRepositoryRepository.cs#L74-L78

source/Octopus.Client/Model/TenantResource.cs
matt-richardson: I tried changing it to:
```diff
-        public string ClonedFromTenantId { get; set; }
+        public string ClonedFromTenantId { get; }
```

and it didn't work... came through as null.

hnrkndrssn: I think if you add the `WriteableOnCreate` attribute it'll make it readonly once created

hnrkndrssn: although, the `ProjectResource` also has a `ClonedFrom` property which is **not** readonly... https://github.com/OctopusDeploy/OctopusClients/blob/master/source/Octopus.Client/Model/ProjectResource.cs#L56

matt-richardson: > I think if you add the `WriteableOnCreate` attribute it'll make it readonly once created

This is done on the server side PR - https://github.com/OctopusDeploy/OctopusDeploy/pull/4285/files#diff-51caff6d0c4dcb6968bdcf3f0ea615ddR70. Not sure it should be done here though?
🤔
Actually, looking through the code - lots of other things have `[WriteableOnCreate]` here... I'm going to stick it on.

source/Octopus.Client/Editors/Async/TenantEditor.cs
```diff
@@ -42,6 +43,12 @@ public async Task<TenantEditor> CreateOrModify(string name)
         
         public async Task<TenantEditor> CreateOrModify(string name, string description, string cloneId = null)
         {
+            var baseRepository = ((TenantRepository) repository).Repository;
```
thedewi: If this component specifically requires a `TenantRepository` instead of any `ITenantRepository`, it should probably depend on that directly. Or this property be exposed in `ITenantRepository`?

source/Octopus.Client/OctopusAsyncRepository.cs
```diff
@@ -176,6 +176,23 @@ public async Task<bool> HasLink(string name)
             var spaceRootDocument = await loadSpaceRootResource.Value.ConfigureAwait(false);
             return spaceRootDocument != null && spaceRootDocument.HasLink(name) || rootDocument.HasLink(name);
         }
+        
+        public async Task<bool> HasLinkParameter(string linkName, string parameterName)
+        {
+            string link;
+            var rootDocument = await loadRootResource.Value.ConfigureAwait(false);
```
thedewi: Is it worth moving this inside the `else` like in the synchronous version, so it only fetches when needed? I don't think it would be too verbose (although some braces might be warranted)

source/Octopus.Client/Editors/Async/TenantEditor.cs
matt-richardson: Agreed this is a bit of a smell.
However, I'm not entirely sure what the best approach is to tackle it... Depending on `TenantRepository` directly cant be done, as all the implementations are internal - only interfaces are exposed. Adding that property to ITenantRepository is also a bit iffy as it exposes stuff that's meant to be internal.
I'm inclined to go "it's not ideal" and move on. I dont think there's much chance this will come back and bite us.

PR #452, opened 2019-08-25
Remove SYNC_CLIENT compile time constant

matt-richardson:
As we now ship a sync client for both net452 and netcore

Also remove `HAS BEST_CONVENTIONAL` as that is also available for netcore.

### DIFF
--- a/source/Octopus.Client.Tests/Integration/Repository/MachineRepositoryTest.cs
+++ b/source/Octopus.Client.Tests/Integration/Repository/MachineRepositoryTest.cs
@@ -41,4 +41,4 @@ namespace Octopus.Client.Tests.Integration.Repository
             Assert.That(tasks.Count, Is.EqualTo(139));
         }
     }
-}    
+}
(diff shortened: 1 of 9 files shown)


source/Octopus.Client.Tests/Integration/Repository/MachineRepositoryTest.cs
```diff
@@ -31,16 +31,14 @@ public void AsyncGetTasksReturnsAllPages()
...
             var repository = new Client.Repositories.MachineRepository(SyncClient.Repository);
             var tasks = repository.GetTasks(machine);
 
             Assert.That(tasks.Count, Is.EqualTo(139));
         }
-#endif
     }
-}
\ No newline at end of file
+}
```
hnrkndrssn: 🔥 whitespace 🔥

matt-richardson:
```suggestion
}
```

Issue #453, opened 2019-08-26
Octopus Client nupkg has dependencies that it shouldn't


gupta-kartik:
https://www.nuget.org/packages/Octopus.Client/ has dependencies of 
* OctoDiff 
* NetwonSoft.JSON

These dependencies should be ILMERGE'd into `Octopus.Client.dll` and not listed in the nupkg file.

PR #454, opened 2019-08-26
ILMerge octodiff, and fix nupkg package to remove unneeded dependencies

matt-richardson:
Because we were using raw `dotnet pack` to create the dll, it was assuming that all the dependencies from the project file should be used as dependencies of the nuget package. However, it didn't know that we'd already ILMerged the `newtonsoft.json` dependency in.

This PR modifies the nupkg so that it no longer lists these dependencies.

We also spotted that octodiff was not being inlined anymore, so this is now being inlined again.

Fixes https://github.com/OctopusDeploy/OctopusClients/issues/453


matt-richardson:
@gupta-kartik tested on both powershell core and powershell fullframework.

PR #458, opened 2019-08-30
Add tests to ensure the packed, ilmerged artifacts work

matt-richardson:
Make sure the packed, ilmerged dll we're going to ship actually works the way we expect it to - ie, dependencies are correct etc

### DIFF
--- a/source/Octopus.Client.E2ETests/NuSpecDependenciesFixture.cs
+++ b/source/Octopus.Client.E2ETests/NuSpecDependenciesFixture.cs
@@ -1,0 +1,190 @@
+using System.IO;
+using System.IO.Compression;
+using System.Linq;
+using System.Reflection;
+using System.Xml;
+using FluentAssertions;
+using NUnit.Framework;
+
+namespace Octopus.Client.E2ETests
+{
+    [TestFixture]
+    public class NuSpecDependenciesFixture
+    {
+        private XmlDocument xmlDoc;
+        private XmlNamespaceManager nameSpaceManager;
+
+        [OneTimeSetUp]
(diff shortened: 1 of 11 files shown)


source/Octopus.Client.E2ETests/NuSpecDependenciesFixture.cs
```diff
@@ -0,0 +1,221 @@
...
+        {
+            var dependency = nuSpecFile.SelectSingleNode(
+                "//ns:package/ns:metadata/ns:dependencies/ns:group[@targetFramework = '.NETStandard2.0']/ns:dependency[@id='Octodiff']",
+                nameSpaceManager);
+            dependency.Should()
+                .NotBeNull("There should be a Octodiff dependency for .NETStandard2.0, as we cant inline it until we get to netcore 3");
+        }
+
+        [Test]
+        public void NetStandard20DependencyOnNewtonsoftJsonShouldBeSameVersionAsCsProj()
```
slewis74: This is why `dotnet pack` puts the csproj dependency versions in the nuspec for you. I bet this is going to catch us out if we have to update the Newtonsoft reference. It'll be up there with the "forgot approval files" commits :)

matt-richardson:
> Question: when one of these tests fails for some reason, will another developer understand why the test exists in the first place, and understand why they probably should keep it running? Or will they be inclined to misunderstand the intent and delete the annoying test?

I'd like to think that we'd be having serious conversations with people if they go round deleting tests without trying to understand the background of a test.

That said, I've added some explanatory xmldoc comments at the top

PR #465, opened 2019-09-25
Replace OpsStepsId reference with the more accurate OpsProcessId

flin-8:


flin-8:
> I think on server we have both properties, so if you're happy to strip this back to just the `OpsProcessId` that sounds good.
> 
> Consider doing the server-side conversion of the resource first, and refactor any references to OpsSteps to confirm on the frontend as well, and that'll give more confidence in the finality of these client resource changes.
> 
> But yea, LGTM

Server side changes are here:
https://github.com/OctopusDeploy/OctopusDeploy/pull/4459

PR #467, opened 2019-09-29
Deployment changes (notes)

slewis74:
New fields to allow setting of a deployment changes template. This is used during the deployment creation to generate ChangesMarkdown, which is used to render in the portal and can be used from email templates etc.

Relates to OctopusDeploy/Issues#5604

### DIFF
--- a/source/Octopus.Client/Model/DeploymentPreviewResource.cs
+++ b/source/Octopus.Client/Model/DeploymentPreviewResource.cs
@@ -10,6 +10,7 @@ namespace Octopus.Client.Model
         public Form Form { get; set; }
         public bool UseGuidedFailureModeByDefault { get; set; }
 
-        public List<ReleaseChanges> ReleaseNotes { get; set; }
+        public List<ReleaseChanges> Changes { get; set; }
+        public string ChangesMarkdown { get; set; }
     }
 }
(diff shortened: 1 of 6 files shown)


source/Octopus.Client/Model/DeploymentPreviewResource.cs
```diff
@@ -6,7 +6,8 @@ namespace Octopus.Client.Model
 {
     public class DeploymentPreviewResource : DeploymentPreviewBaseResource
     {
-        public List<ReleaseChanges> ReleaseNotes { get; set; }
+        public List<ReleaseChanges> Changes { get; set; }
```
LittleBigMoosey: This breaking change corresponds with the changes in Server for this resource 👍

source/Octopus.Client/Model/DeploymentProcessResource.cs
```diff
@@ -3,6 +3,7 @@
 using System.ComponentModel.DataAnnotations;
 using System.Linq;
 using Octopus.Client.Extensibility;
+using Octopus.Client.Extensibility.Attributes;
```
hnrkndrssn: This looks like an unintended addition

source/Octopus.Client/Model/DeploymentResource.cs
```diff
@@ -74,7 +76,7 @@ public DeploymentBaseResource()
 
         [WriteableOnCreate]
         public string Comments { get; set; }
-
+
```
hnrkndrssn: 🔥 white space 🔥

PR #469, opened 2019-10-03
Dynamic Worker Health status

benPearce1:


benPearce1:
[How would we know that deployments are failing because DW can't be provisioned.](https://trello.com/c/0A6rpYH6/585-how-would-we-know-that-deployments-are-failing-because-dw-cant-be-provisioned)

benPearce1:
No longer required

PR #474, opened 2019-10-29
WIP - Client update for added project trigger enhancements

alialmaktoum:
@ankithkonda and @alialmaktoum  working on Octopus Client changes for the following batch of work -

**Private links**
https://trello.com/c/TMROpfwR/79-add-description-field-to-triggers
https://trello.com/c/We1E5bg8/80-add-end-time-field-provides-symmetry-for-start-time
https://trello.com/c/Lsi8NJIi/74-run-after-and-run-until
Server changes at: https://github.com/OctopusDeploy/OctopusDeploy/pull/4655


NickJosevski:
### Breaking Change

We carefully considered the impact of this breaking change. Our decission to break the API and resources here is that we are classifying the previous behaviour as confusing and can lead to misconfiguration.

We are introducing client and server checks to help guild this breaking change, but it is a hard break.
 - A new client can no longer communicate with old server to automate scheduled triggers (code has changed).
 - An old client will get human readable errors when contacing a new server (post 2019.11) that they cannot automate triggers anymore from that version.

PR #478, opened 2019-11-05
Support fetching the Release Log.

thedewi:
Adds and implements `IReleaseRepository.GetLog()`.

Part of a solution to OctopusDeploy/Issues#5869.

### DIFF
--- a/source/Octopus.Client/Repositories/ReleaseRepository.cs
+++ b/source/Octopus.Client/Repositories/ReleaseRepository.cs
@@ -1,4 +1,5 @@
 using System;
+using Octopus.Client.Exceptions;
 using Octopus.Client.Model;
 
 namespace Octopus.Client.Repositories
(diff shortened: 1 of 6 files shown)


source/Octopus.Client/Repositories/ReleaseRepository.cs
```diff
@@ -70,5 +71,10 @@ public LifecycleProgressionResource GetProgression(ReleaseResource release)
         {
             return Client.Get<LifecycleProgressionResource>(release.Links["Progression"]);
         }
+
+        public ResourceCollection<ReleaseLogResource> GetLog(ReleaseResource release, int skip = 0, int? take = null)
+        {
```
slewis74: This should also have the check for the "Log" link and the message about the required server version

thedewi: Okay done.

PR #482, opened 2019-11-19
Backwards Compat Check

NickJosevski:
For the cases where we have decided we must make a breaking change in Server and Octopus.Client.

The client will have a resource that's no longer compatible with *old versions of server*. We want to alert/protect the customer in this scenario.

- We want this check to only happen at the time of calling specific API (method) on specific repositories (not at repository creation time).
- It may often impact the entire repository.
   - This is the case with this most recent upcoming change.
   - This is also the case with the historic  Team/ScopedUserRole/UserRole change made for 2019.1.0

Here is the start of this idea.

### Future breaking changes to the same resources/repositories
 - Would be fine, as the `versionThreshold` would just be bumped up higher again.

### Results

Using client against:
 1. Server 2019.10 no error
 2. Server 2018.10 expected to see error

### Sync client
image

### Async client
image

### DIFF
--- a/source/Octopus.Client/Repositories/Async/BasicRepository.cs
+++ b/source/Octopus.Client/Repositories/Async/BasicRepository.cs
@@ -1,7 +1,9 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Runtime.CompilerServices;
 using System.Text.RegularExpressions;
+using System.Threading;
 using System.Threading.Tasks;
 using Octopus.Client.Exceptions;
 using Octopus.Client.Extensibility;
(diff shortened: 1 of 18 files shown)


source/Octopus.Client/Repositories/Async/BasicRepository.cs
```diff
@@ -55,14 +58,44 @@ protected virtual void CheckSpaceResource(IHaveSpaceResource spaceResource)
...
+
+            var currentServerVersion = await CurrentVersion();
+
+            if (currentServerVersion.Some() && currentServerVersion.Value < versionThreshold)
+            {
+                throw new NotSupportedException(
+                    $"The version of the Octopus Server ('{currentServerVersion.Value}') you are connecting to is not compatible with this version of Octopus.Client for this API call. Please upgrade your Octopus Server to a version greater than '{versionThreshold}'");
+            }
+            
+            async Task<Maybe<SemanticVersion>> CurrentVersion()
```
flin-8: Consider adding a retry to this to see if that makes the race condition go away

source/Octopus.Client/Repositories/Async/BasicRepository.cs
```diff
@@ -55,14 +58,44 @@ protected virtual void CheckSpaceResource(IHaveSpaceResource spaceResource)
                 });
         } 
 
+        protected void SetBackwardsCompatVersionThreshold(string version)
```
flin-8: Rename to `MinimumCompatibleVersion`

PR #484, opened 2019-11-21
Add project trigger filtering by runbook

tothegills:
Provide runbook ids to the method `FindByRunbook` and the result will be any project triggers matching those ids.


tothegills:
[Filter triggers by runbook](https://trello.com/c/lU33A0cC/98-filter-triggers-by-runbook)

PR #487, opened 2019-11-27
Improve development experience in non-windows environments

TomPeters:
This PR makes it possible to do basic development of this library in non-windows OSes like linux or macOS. Only the compatible target frameworks are compiled in those OSes.


source/Octopus.Client.E2ETests/TestHelpers.cs
```diff
@@ -10,7 +10,7 @@ public static class TestHelpers
         internal static string GetNuGetPackage()
         {
             var path = Path.GetDirectoryName(Assembly.GetExecutingAssembly().FullLocalPath());
-            var artifactsFolder = Path.GetFullPath(Path.Combine(path, @"..\..\..\..\..\artifacts"));
+            var artifactsFolder = Path.GetFullPath(Path.Combine(path, "..", "..", "..", "..", "..", "artifacts"));
```
jburger: Good one to look out for in future :+1:

PR #497, opened 2019-12-05
Don't assume the file hasn't been uploaded

pawelpabich:
Fixes OctopusDeploy/Issues#6055

It takes a long time to upload a large (< 2GB) file to Hosted v1 and Hosted v2. The data gets persisted relatively quickly but the client, in this case, otcto.exe,  doesn't get notified about this and waits until the call times out. 

**Before**

Data uploaded successfully but the client throws `TimeoutException`.

**Now**

We still wait until the timeout exception but this time the overall command succeeds.

image

image

When the client gets `TimeoutException` it checks whether the package has been uploaded successfully.  If it has, it doesn't return an error.

We still don't know what the causes the delay but this change at least helps our Cloud customers to work around the problem.  If you compare the time when the client finished and the time when the package was uploaded you will see a gap of 17 and 27 minutes.

This change also makes sense in isolation because we really don't know whether an operation was successful when it times out.

### DIFF
--- a/source/Octopus.Client/Repositories/Async/BuiltInPackageRepositoryRepository.cs
+++ b/source/Octopus.Client/Repositories/Async/BuiltInPackageRepositoryRepository.cs
@@ -23,7 +23,7 @@ namespace Octopus.Client.Repositories.Async
         Task DeletePackage(PackageResource package);
         Task DeletePackages(IReadOnlyList<PackageResource> packages);
 
-        Task<PackageFromBuiltInFeedResource> GetPackage(string packageId);
+        Task<PackageFromBuiltInFeedResource> GetPackage(string packageId, string version);
     }
 
     class BuiltInPackageRepositoryRepository : IBuiltInPackageRepositoryRepository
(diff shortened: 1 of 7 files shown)


source/Octopus.Client/Repositories/Async/BuiltInPackageRepositoryRepository.cs
```diff
@@ -59,6 +61,13 @@ public async Task<PackageFromBuiltInFeedResource> PushPackage(string fileName, S
                     if (deltaResult != null)
                         return deltaResult;
                 }
+                catch (TimeoutException ex)
+                {
+                    Logger.Info("Delta push timed out: " + ex.Message);
+
+                    var verificationResult = await VerifyTransfer(fileName, contents);
```
droyad: This is missing `.ConfigureAwait(false)`

source/Octopus.Client/Repositories/Async/BuiltInPackageRepositoryRepository.cs
```diff
@@ -85,14 +94,49 @@ public async Task<PackageFromBuiltInFeedResource> PushPackage(string fileName, S
...
+            var uploadedPackage = await TryFindPackage(packageId, version);
+
+            return PackageContentComparer.AreSame(uploadedPackage, contents, Logger) ? uploadedPackage : null;
+        }
+
+        private async Task<PackageFromBuiltInFeedResource> TryFindPackage(string packageId, SemanticVersion version)
+        {
+            try
+            {
+                return await repository.BuiltInPackageRepository.GetPackage($"{packageId}.{version}");
```
droyad: `.ConfigureAwait(false)`

pawelpabich:
@slewis74 👍 I simply exposed the raw Ocotpus API. Can you have a look again?

PR #501, opened 2020-01-08
Bug osx ignore ssl errors

matt-richardson:


matt-richardson:
> Curious about the `StringComparer.OrdinalIgnoreCase`. I assume that's what changed the order in the approval files. I thought the order made more sense the way it was.

It was all working fine on all the other build agents except ubuntu18 - it was less about getting a desirable order, and more about get a consistent order. Not sure why ubuntu18 thought it was special.

PR #506, opened 2020-01-29
Bug in OctopusClients prevents it from being used with non-space supported servers

alialmaktoum:
OctopusClient needs to be updated to handle connecting to a server that does not know about spaces

Internal links:
https://trello.com/c/W7Tj8qJI/2675-bug-in-octopusclients-prevents-it-from-being-used-with-non-space-supported-servers

### DIFF
--- a/source/Octopus.Client/Repositories/Async/BasicRepository.cs
+++ b/source/Octopus.Client/Repositories/Async/BasicRepository.cs
@@ -51,7 +51,7 @@ namespace Octopus.Client.Repositories.Async
                     var spaceRoot = await Repository.LoadSpaceRootDocument().ConfigureAwait((false));
                     var isDefaultSpaceFound = spaceRoot != null;
 
-                    if (!isDefaultSpaceFound && await ServerSupportsSpaces())
+                    if (!isDefaultSpaceFound && await ServerSupportsSpaces().ConfigureAwait(false))
                     {
                         throw new DefaultSpaceNotFoundException(spaceResource);
                     }
--- a/source/Octopus.Client/Repositories/Async/BasicRepository.cs
+++ b/source/Octopus.Client/Repositories/Async/BasicRepository.cs
@@ -21,7 +21,8 @@ namespace Octopus.Client.Repositories.Async
         private SemanticVersion minimumRequiredVersion;
         private bool hasMinimumRequiredVersion;
 
-        protected BasicRepository(IOctopusAsyncRepository repository, string collectionLinkName, Func<IOctopusAsyncRepository, Task<string>> getCollectionLinkName = null)
+        protected BasicRepository(IOctopusAsyncRepository repository, string collectionLinkName,
+            Func<IOctopusAsyncRepository, Task<string>> getCollectionLinkName = null)
         {
             Client = repository.Client;
             Repository = repository;
@@ -46,25 +47,25 @@ namespace Octopus.Client.Repositories.Async
                         throw new ResourceSpaceDoesNotMatchRepositorySpaceException(spaceResource, space);
                 },
                 whenSystemScoped: () => { },
-                whenUnspecifiedScope: () =>
+                whenUnspecifiedScope: async () =>
                 {
-                    var spaceRoot = Repository.LoadSpaceRootDocument().Result;
+                    var spaceRoot = await Repository.LoadSpaceRootDocument().ConfigureAwait((false));
                     var isDefaultSpaceFound = spaceRoot != null;
-                    
-                    if (!isDefaultSpaceFound && CheckServerVersionIncludesSpaces(Repository.LoadRootDocument().Result.Version))
+
+                    if (!isDefaultSpaceFound && await ServerSupportsSpaces())
                     {
-                            throw new DefaultSpaceNotFoundException(spaceResource);
+                        throw new DefaultSpaceNotFoundException(spaceResource);
                     }
                 });
         }
-        
-        private bool CheckServerVersionIncludesSpaces(string version)
-        {
-            var versionOfServer = SemanticVersion.Parse(version);
-            var versionSpacesIntroduced = SemanticVersion.Parse("2019.1.0");
-            var versionIncludesSpaces = versionOfServer >= versionSpacesIntroduced;
 
-            return versionIncludesSpaces;
+        private async Task<bool> ServerSupportsSpaces()
+        {
+            var rootDocument = await Repository.LoadRootDocument().ConfigureAwait(false);
+
+            var spacesIsSupported = rootDocument.HasLink("Spaces");
+
+            return spacesIsSupported;
         }
 
         protected void MinimumCompatibleVersion(string version)
@@ -120,7 +121,8 @@ namespace Octopus.Client.Repositories.Async
             return Client.Delete(resource.Links["Self"]);
         }
 
-        public async Task Paginate(Func<ResourceCollection<TResource>, bool> getNextPage, string path = null, object pathParameters = null)
+        public async Task Paginate(Func<ResourceCollection<TResource>, bool> getNextPage, string path = null,
+            object pathParameters = null)
         {
             await ThrowIfServerVersionIsNotCompatible();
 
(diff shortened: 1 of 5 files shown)


alialmaktoum:
[Bug in OctopusClients prevents it from being used with non-space supported servers](https://trello.com/c/W7Tj8qJI/2675-bug-in-octopusclients-prevents-it-from-being-used-with-non-space-supported-servers)

source/Octopus.Client/Repositories/BasicRepository.cs
```diff
@@ -47,13 +47,22 @@ protected virtual void CheckSpaceResource(IHaveSpaceResource spaceResource)
...
-                    if (!isDefaultSpaceFound)
+                  
+                    if (!isDefaultSpaceFound && CheckServerVersionIncludesSpaces(Repository.LoadRootDocument().Version))
                     {
                         throw new DefaultSpaceNotFoundException(spaceResource);
                     }
                 });
         }
+        
+        private bool CheckServerVersionIncludesSpaces(string version)
```
TomPeters: While this implementation works, it might be better to do feature detection, rather than hard coding versions. We already do this over here: https://github.com/OctopusDeploy/OctopusClients/blob/master/source/Octopus.Client/OctopusAsyncRepository.cs#L295

TomPeters: [nit] A better name for this function might be something like `ServerSupportsSpaces` or `IsSpacesSupported` or `IsSpacesSupportedByServer`. In particular, the `Check` part of the name isn't very helpful. The method name would be better if it only reflected the value we are trying to obtain.

source/Octopus.Client/Repositories/Async/BasicRepository.cs
```diff
@@ -48,15 +48,24 @@ protected virtual void CheckSpaceResource(IHaveSpaceResource spaceResource)
                 whenSystemScoped: () => { },
                 whenUnspecifiedScope: () =>
                 {
-                    var spaceRoot = Repository.LoadSpaceRootDocument();
+                    var spaceRoot = Repository.LoadSpaceRootDocument().Result;
```
TomPeters: We can't use `.Result` here. While this would work in applications like ours, this could introduced deadlocks in other applications that use a different `SynchronizationContext` (eg winforms or .net framework asp.net). The correct way to do this is `var spaceRoot = await Repository.LoadSpaceRootDocument().ConfigureAwait(false);`.

The `.ConfigureAwait(false)` part is also really important for the same reason (prevents deadlocks). This is a really common pattern in async code like this that exists in libraries (if you search for this in this repo, you should find it everywhere we use `await`), but is something that we would tend to omit from code that we don't publish as libraries (eg in our OctopusServer repo).

In fact, this highlights an existing bug - the current implementation doesn't actually work. The line below where it goes `spaceRoot != null` is actually check whether the `Task` is null (which would never happen), rather than the `Task`s result.

The problem here is that you need to keep making everything `async` all the way up to get this to work, so this change will affect at least a few more methods, but I don't think its too much more work in this case.

Let me know if you need a hand with this.

source/Octopus.Client/Repositories/BasicRepository.cs
```diff
@@ -47,13 +47,22 @@ protected virtual void CheckSpaceResource(IHaveSpaceResource spaceResource)
                 {
                     var spaceRoot = Repository.LoadSpaceRootDocument();
                     var isDefaultSpaceFound = spaceRoot != null;
-
-                    if (!isDefaultSpaceFound)
+                  
+                    if (!isDefaultSpaceFound && CheckServerVersionIncludesSpaces(Repository.LoadRootDocument().Version))
```
TomPeters: Unlike the async version, this one is fine :+1:

source/Octopus.Client/Repositories/Async/BasicRepository.cs
```diff
@@ -46,18 +46,27 @@ protected virtual void CheckSpaceResource(IHaveSpaceResource spaceResource)
                         throw new ResourceSpaceDoesNotMatchRepositorySpaceException(spaceResource, space);
                 },
                 whenSystemScoped: () => { },
-                whenUnspecifiedScope: () =>
+                whenUnspecifiedScope: async () =>
```
TomPeters: This isn't quite right yet - the caller doesn't await the result of this, so the exception may be ignored in some cases.

Feel free to grab me to help you resolve this over zoom. I may be difficult to get a hold of next week - you may have more success getting help from someone else in QRF that is good at `async/await`.

source/Octopus.Client/Repositories/Async/BasicRepository.cs
```diff
@@ -37,25 +37,35 @@ protected BasicRepository(IOctopusAsyncRepository repository, string collectionL
...
-        protected virtual void CheckSpaceResource(IHaveSpaceResource spaceResource)
+        protected virtual async Task CheckSpaceResource(IHaveSpaceResource spaceResource)
         {
-            Repository.Scope.Apply(
+            await Repository.Scope.Apply(
                 whenSpaceScoped: space =>
                 {
                     if (spaceResource.SpaceId != null && spaceResource.SpaceId != space.Id)
                         throw new ResourceSpaceDoesNotMatchRepositorySpaceException(spaceResource, space);
+                    return Task.FromResult("");
```
TomPeters: [nit] I believe it is more common to do `Task.FromResult(null)`

source/Octopus.Client/Repositories/Async/BasicRepository.cs
```diff
@@ -37,25 +37,35 @@ protected BasicRepository(IOctopusAsyncRepository repository, string collectionL
         public IOctopusAsyncClient Client { get; }
         public IOctopusAsyncRepository Repository { get; }
 
-        protected virtual void CheckSpaceResource(IHaveSpaceResource spaceResource)
+        protected virtual async Task CheckSpaceResource(IHaveSpaceResource spaceResource)
         {
-            Repository.Scope.Apply(
+            await Repository.Scope.Apply(
```
TomPeters: [nit] You could just return the promise here instead

`return Repository.Scope.Apply(...)` + the method does not need to be `async`

This has the advantage of not needing the `.ConfigureAwait(false)`, because you have not done an `await`!

source/Octopus.Client/Repositories/Async/MixedScopeBaseRepository.cs
```diff
@@ -41,16 +42,18 @@ void ValidateThatICanUseACustomSpaceContext()
             }
         }
 
-        protected override void CheckSpaceResource(IHaveSpaceResource spaceResource)
+        protected override async Task CheckSpaceResource(IHaveSpaceResource spaceResource)
         {
-            Repository.Scope.Apply(
+            await Repository.Scope.Apply(
```
TomPeters: [nit]

Same as above - could just `return` the task directly instead of awaiting.

source/Octopus.Client/Repositories/Async/BasicRepository.cs
alialmaktoum: `Task.FromResult(null)` would only work as `Task.FromResult<object>(null)`. I have used another common approach `Task.FromResult(0)`

source/Octopus.Client/Repositories/Async/MixedScopeBaseRepository.cs
```diff
@@ -41,16 +42,18 @@ void ValidateThatICanUseACustomSpaceContext()
             }
         }
 
-        protected override void CheckSpaceResource(IHaveSpaceResource spaceResource)
+        protected override Task CheckSpaceResource(IHaveSpaceResource spaceResource)
         {
-            Repository.Scope.Apply(
+            return Repository.Scope.Apply(
```
TomPeters: Actually this can be simplified by returning the task at the end

```
protected override Task CheckSpaceResource(IHaveSpaceResource spaceResource)
{
    Repository.Scope.Apply(
        whenSpaceScoped: space =>
                {
                    if (spaceResource.SpaceId != null && spaceResource.SpaceId != space.Id)
                        throw new ResourceSpaceDoesNotMatchRepositorySpaceException(spaceResource, space);
                },
                whenSystemScoped: () => { },
                whenUnspecifiedScope: () => { });
    return Task.FromResult(null);
}
```

PR #510, opened 2020-03-03
Add xFrameOptions configure option and deprecate xFrameOptionAllowFrom

alialmaktoum:
# Background
xFrameOptionAllowFrom has been marked as deprecated as allow-from has been deprecated: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Frame-Options

new option XFrameOptions added which allows customers to set x-frame-options header to sameorigin, deny etc.

https://trello.com/c/qMkihfxa/2944-x-frame-options
Fixes https://github.com/OctopusDeploy/Issues/issues/6196

### DIFF
--- a/source/Octopus.Client.Tests/PublicSurfaceAreaFixture.ThePublicSurfaceAreaShouldNotRegress..NETCore.approved.txt
+++ b/source/Octopus.Client.Tests/PublicSurfaceAreaFixture.ThePublicSurfaceAreaShouldNotRegress..NETCore.approved.txt
@@ -5222,8 +5222,10 @@ Octopus.Client.Model
   class XOptionsResource
   {
     static System.String XFrameAllowFromDescription
+    static System.String XFrameOptionsDescription
     .ctor()
     String XFrameOptionAllowFrom { get; set; }
+    String XFrameOptions { get; set; }
   }
 }
 Octopus.Client.Model.Accounts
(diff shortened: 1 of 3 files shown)


alialmaktoum:
[X-Frame-Options](https://trello.com/c/qMkihfxa/2944-x-frame-options)

source/Octopus.Client.Tests/PublicSurfaceAreaFixture.ThePublicSurfaceAreaShouldNotRegress..NETCore.approved.txt
```diff
@@ -5222,8 +5222,10 @@ Octopus.Client.Model
   class XOptionsResource
   {
     static System.String XFrameAllowFromDescription
+    static System.String XFrameOptionsDescription
```
matt-richardson: Do these need to be public? Seems... odd.

alialmaktoum: yes does not make sense to me either.. i will remove it

alialmaktoum: I think it was done this way because XOptionsResource on server also has the descriptions as public

PR #514, opened 2020-03-12
Add a deprecation notice when client is being used from server's folder

slewis74:
This causes trouble with NETFramework vs NETCore, especially in PowerShell 5.

### DIFF
--- a/source/Octopus.Client/LocationChecker.cs
+++ b/source/Octopus.Client/LocationChecker.cs
@@ -24,10 +24,10 @@ namespace Octopus.Client
             var applicationRegNode = octopusRegNode.OpenSubKey(application);
             if (applicationRegNode != null)
             {
-                var serverInstallationFolder = (string) applicationRegNode.GetValue("InstallLocation");
+                var installationFolder = (string) applicationRegNode.GetValue("InstallLocation");
                 var currentAssemblyLocation = typeof(LocationChecker).Assembly.Location;
 
-                if (currentAssemblyLocation.Contains(serverInstallationFolder))
+                if (currentAssemblyLocation.Contains(installationFolder))
                 {
                     var warningMessage = $"Using Octopus.Client from the {application}'s installation folder has been deprecated. In future versions it may not be shipped with {application}.";
                     
--- a/source/Octopus.Client/LocationChecker.cs
+++ b/source/Octopus.Client/LocationChecker.cs
@@ -14,26 +14,31 @@ namespace Octopus.Client
 
         public static void CheckAssemblyLocation()
         {
-            var octopusRegNode = Registry.LocalMachine.OpenSubKey(@"Software\Octopus");
-            CheckBasedOnApplication(octopusRegNode, "Server");
-            CheckBasedOnApplication(octopusRegNode, "Tentacle");
+            using (var octopusRegNode = Registry.LocalMachine.OpenSubKey(@"Software\Octopus"))
+            {
+                CheckBasedOnApplication(octopusRegNode, "Server");
+                CheckBasedOnApplication(octopusRegNode, "Tentacle");
+            }
         }
 
         private static void CheckBasedOnApplication(RegistryKey octopusRegNode, string application)
         {
-            var applicationRegNode = octopusRegNode.OpenSubKey(application);
-            if (applicationRegNode != null)
+            using (var applicationRegNode = octopusRegNode.OpenSubKey(application))
             {
-                var installationFolder = (string) applicationRegNode.GetValue("InstallLocation");
-                var currentAssemblyLocation = typeof(LocationChecker).Assembly.Location;
-
-                if (currentAssemblyLocation.Contains(installationFolder))
+                if (applicationRegNode != null)
                 {
-                    var warningMessage = $"Using Octopus.Client from the {application}'s installation folder has been deprecated. In future versions it may not be shipped with {application}. Please see https://g.octopushq.com/usingoctopusclientfrominstallationfolder.";
-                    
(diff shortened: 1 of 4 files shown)


source/Octopus.Client/LocationChecker.cs
```diff
@@ -0,0 +1,34 @@
+using System;
+using System.Diagnostics;
+using System.Linq;
+using System.Reflection;
+using Microsoft.Win32;
+using Octopus.Client.Logging;
+
+namespace Octopus.Client
+{
+#if FULL_FRAMEWORK
```
matt-richardson: if you move this down inside the `CheckAssemblyLocation` method, you can ditch the `#if`s in the other locations

source/Octopus.Client/LocationChecker.cs
```diff
@@ -0,0 +1,34 @@
...
+            var octopusRegNode = Registry.LocalMachine.OpenSubKey(@"Software\Octopus");
+            var serverRegNode = octopusRegNode.OpenSubKey("Server");
+            if (serverRegNode != null)
+            {
+                var serverInstallationFolder = (string)serverRegNode.GetValue("InstallLocation");
+                var currentAssemblyLocation = typeof(LocationChecker).Assembly.Location;
+
+                if (currentAssemblyLocation.Contains(serverInstallationFolder))
+                {
+                    var warningMessage = $"Using Octopus.Client from the server's installation folder has been deprecated. In future versions it may not be shipped with server.";
```
matt-richardson:
```suggestion
                    var warningMessage = $"Using Octopus.Client from the server's installation folder has been deprecated. In future versions it may not be shipped with server. Please see https://g.octopushq.com/usingoctopusclientfrominstallationfolder.";
```

source/Octopus.Client/LocationChecker.cs
```diff
@@ -0,0 +1,34 @@
...
+{
+#if FULL_FRAMEWORK
+    class LocationChecker
+    {
+        private static readonly ILog Logger = LogProvider.For<LocationChecker>();
+
+        public static void CheckAssemblyLocation()
+        {
+            var octopusRegNode = Registry.LocalMachine.OpenSubKey(@"Software\Octopus");
+            var serverRegNode = octopusRegNode.OpenSubKey("Server");
```
matt-richardson: You decided against worrying about Tentacle?

source/Octopus.Client/LocationChecker.cs
slewis74: I decided to go this way so the whole thing just doesn't exist in the NETCore world

source/Octopus.Client/LocationChecker.cs
slewis74: I did. Could Tentacle be calling it from the install folder from inside PowerShell? 🤔 Would this happen if you ran a script step on a worker?

matt-richardson: I think the only way this would happen would be if it was a customer script, so we'd want the warning there too

slewis74: Updated to include warning for Tentacle too.

source/Octopus.Client/LocationChecker.cs
johnsimons: @slewis74 you could move the conditional compile directive to the top of the file

source/Octopus.Client/LocationChecker.cs
```diff
@@ -0,0 +1,42 @@
...
+namespace Octopus.Client
+{
+#if FULL_FRAMEWORK
+    class LocationChecker
+    {
+        private static readonly ILog Logger = LogProvider.For<LocationChecker>();
+
+        public static void CheckAssemblyLocation()
+        {
+            var octopusRegNode = Registry.LocalMachine.OpenSubKey(@"Software\Octopus");
```
johnsimons: I think `octopusRegNode` instance is `IDisposable`?

source/Octopus.Client/LocationChecker.cs
```diff
@@ -0,0 +1,42 @@
...
+            CheckBasedOnApplication(octopusRegNode, "Server");
+            CheckBasedOnApplication(octopusRegNode, "Tentacle");
+        }
+
+        private static void CheckBasedOnApplication(RegistryKey octopusRegNode, string application)
+        {
+            var applicationRegNode = octopusRegNode.OpenSubKey(application);
```
johnsimons: `applicationRegNode` is `IDisposable`

source/Octopus.Client/LocationChecker.cs
```diff
@@ -0,0 +1,42 @@
...
+            if (applicationRegNode != null)
+            {
+                var installationFolder = (string) applicationRegNode.GetValue("InstallLocation");
+                var currentAssemblyLocation = typeof(LocationChecker).Assembly.Location;
+
+                if (currentAssemblyLocation.Contains(installationFolder))
+                {
+                    var warningMessage = $"Using Octopus.Client from the {application}'s installation folder has been deprecated. In future versions it may not be shipped with {application}. Please see https://g.octopushq.com/usingoctopusclientfrominstallationfolder.";
```
johnsimons: "It seems your script is referencing Octopus.Client.dll from the {application}'s installation folder. This is a heads up to let you know that in future releases Octopus.Client.dll will no longer be shipped with {application}. Please refer to https://g.octopushq.com/usingoctopusclientfrominstallationfolder for instructions on how to address this issue.

source/Octopus.Client/LocationChecker.cs
slewis74: Indeed it is, good 👀 ! Fixed, and the other one below too

PR #515, opened 2020-03-12
Bring ScopeField enum into sync with server

matt-richardson:
To fix customers who might hit deserialization errors if the enum is sent from the server


matt-richardson:
> I wonder if this would have been nicer/easier to read if the additional enum value was just added to the end? Doesn't matter to the end result of the executing code, just a thought on making the PR easier to read.

Agreed, but thought i'd take the hit on this one and do a full copy and paste from server, so it's identical.

PR #517, opened 2020-03-23
Revert "Action Container Images Model Changes"

alialmaktoum:
Reverts OctopusDeploy/OctopusClients#516


droyad:
[Investigate script source equal to package rather than inline](https://trello.com/c/etfsPnGl/53-investigate-script-source-equal-to-package-rather-than-inline)

PR #536, opened 2020-07-06
Add includePrevious arg to dashboard repository dynamic dashboard

shaunmarx:
Add `includePrevious` arg to dashboard repository dynamic dashboard

Related: 
https://github.com/OctopusDeploy/OctopusCLI/pull/94
https://github.com/OctopusDeploy/OctopusCLI/pull/94

### DIFF
--- a/source/Octopus.Client.Tests/PublicSurfaceAreaFixture.ThePublicSurfaceAreaShouldNotRegress..NETCore.approved.txt
+++ b/source/Octopus.Client.Tests/PublicSurfaceAreaFixture.ThePublicSurfaceAreaShouldNotRegress..NETCore.approved.txt
@@ -2308,6 +2308,11 @@ Octopus.Client.Model
     String TaskId { get; set; }
     String TenantId { get; set; }
   }
+  DashboardItemsOptions
+  {
+      IncludeCurrentDeploymentOnly = 0
+      IncludeCurrentAndPreviousSuccessfulDeployment = 1
+  }
   class DashboardProjectGroupResource
     Octopus.Client.Extensibility.IResource
     Octopus.Client.Model.IAuditedResource
@@ -6618,7 +6623,7 @@ Octopus.Client.Repositories
   interface IDashboardRepository
   {
     Octopus.Client.Model.DashboardResource GetDashboard()
-    Octopus.Client.Model.DashboardResource GetDynamicDashboard(String[], String[], Boolean)
+    Octopus.Client.Model.DashboardResource GetDynamicDashboard(String[], String[], Octopus.Client.Model.DashboardItemsOptions)
   }
   interface IDefectsRepository
   {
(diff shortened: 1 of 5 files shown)


ankithkonda:
[Octo CLI not using latest release, not latest successful release for deploy and promotion](https://trello.com/c/UiBuRXJr/3380-octo-cli-not-using-latest-release-not-latest-successful-release-for-deploy-and-promotion)

source/Octopus.Client.Tests/PublicSurfaceAreaFixture.ThePublicSurfaceAreaShouldNotRegress..NETCore.approved.txt
```diff
@@ -6618,7 +6618,7 @@ Octopus.Client.Repositories
   interface IDashboardRepository
   {
     Octopus.Client.Model.DashboardResource GetDashboard()
-    Octopus.Client.Model.DashboardResource GetDynamicDashboard(String[], String[])
+    Octopus.Client.Model.DashboardResource GetDynamicDashboard(String[], String[], Boolean)
```
tothegills, line 6621: What happens to customers who are currently using `GetDynamicDashboard`?

ankithkonda: Ah yes! good point. um would an overload method be solution to that?

tothegills: Yes, I think an overloaded method would be a good solution or a default value in the interface.

source/Octopus.Client/Repositories/DashboardRepository.cs
```diff
@@ -6,9 +6,9 @@ namespace Octopus.Client.Repositories
     public interface IDashboardRepository
     {
         DashboardResource GetDashboard();
-        DashboardResource GetDynamicDashboard(string[] projects, string[] environments);
+        DashboardResource GetDynamicDashboard(string[] projects, string[] environments, bool includePrevious = false);
```
slewis74, line 9: What does this param actually do? As a caller it isn't clear to me what previous thing is actually going to get included here if I pass `true`.

ankithkonda: Ah so right now, the DashboardResource has an `items` property which includes a list of releases, However as it is right now, it only ever gets you 1 item which is the latest release. The dashboard route however, does let you set this to true, and therefore returns all the previous releases that were successful. We need this in OctopusCLI in order to promote the most recent successful release.

But yes! it's not really clear. Perhaps instead of a straight boolean, 
So I guess two options - 

1) add documentation to that method, explaining what the parameter is for

2) add DynamicDashboardOptions type, which has more descriptive `includePreviousReleases` property? 

```
DashboardResource GetDynamicDashboard(string[] projects, string[] environments, DynamicDashboardOptions options);

...
GetDynamicDashboard(projects, environments, {includePreviousOptions:true})

```

tothegills: Even wrapped up in a type, what does `includePreviousReleases:true` mean?

PR #538, opened 2020-07-21
Adds beta config-as-code endpoints

andrewabest:
Background
===

As a part of the Configuration As Code: Multi Branch Round Trip pitch, we will be implementing the ability to select a branch within the Octopus user interface.

This PR adds client methods for endpoints created in https://github.com/OctopusDeploy/OctopusDeploy/pull/6444 and https://github.com/OctopusDeploy/OctopusDeploy/pull/6408

These methods will be used to create e2e tests for the server application.

**Do not merge** until https://github.com/OctopusDeploy/OctopusDeploy/pull/6444 is merged

### DIFF
--- a/source/Octopus.Client/Repositories/Async/DeploymentProcessRepository.cs
+++ b/source/Octopus.Client/Repositories/Async/DeploymentProcessRepository.cs
@@ -35,8 +35,7 @@ namespace Octopus.Client.Repositories.Async
 
     public interface IDeploymentProcessRepositoryBeta
     {
-        Task<DeploymentProcessResource> Get(ProjectResource projectResource, string gitref);
-        Task Modify(DeploymentProcessResource deploymentProcessResource, ProjectResource projectResource, string gitref);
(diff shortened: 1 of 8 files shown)


source/Octopus.Client/Repositories/Async/DeploymentProcessRepository.cs
```diff
@@ -6,19 +6,59 @@ namespace Octopus.Client.Repositories.Async
...
+
         public DeploymentProcessRepository(IOctopusAsyncRepository repository)
             : base(repository, "DeploymentProcesses")
         {
+            beta = new DeploymentProcessRepositoryBeta(repository);
+        }
+
+        public IDeploymentProcessRepositoryBeta Beta(bool useBeta)
+        {
+            if (!useBeta) throw new Exception($"You must supply true for {nameof(useBeta)} to use Beta functionality.");
```
droyad, line 25: This is likely overkill, it's obvious to me from the property name that it's beta and I'm opting in to ☠️

PR #540, opened 2020-09-18
Support compression

adam-mccoy:
# Background

The client library doesn't take advantage of compression features of HTTP by sending the `Accept-Encoding` header. This PR sets the `AutomaticDecompression` option, which will handle setting the appropriate header and performing the decompression of the result stream automatically.

### DIFF
--- a/source/Octopus.Client/OctopusAsyncClient.cs
+++ b/source/Octopus.Client/OctopusAsyncClient.cs
@@ -46,6 +46,11 @@ namespace Octopus.Client
                 Credentials = serverEndpoint.Credentials ?? CredentialCache.DefaultNetworkCredentials,
             };
 
+            if (handler.SupportsAutomaticDecompression)
+            {
+                handler.AutomaticDecompression = DecompressionMethods.Deflate | DecompressionMethods.GZip;
+            }
+
             if (clientOptions.Proxy != null)
             {
                 handler.UseProxy = true;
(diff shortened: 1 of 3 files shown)


adam-mccoy:
[Client Lib does not support compression](https://trello.com/c/6iTJgHwx/3710-client-lib-does-not-support-compression)

source/Octopus.Client/OctopusAsyncClient.cs
```diff
@@ -46,6 +46,11 @@ protected OctopusAsyncClient(OctopusServerEndpoint serverEndpoint, OctopusClient
                 Credentials = serverEndpoint.Credentials ?? CredentialCache.DefaultNetworkCredentials,
             };
 
+            if (handler.SupportsAutomaticDecompression)
+            {
+                handler.AutomaticDecompression = DecompressionMethods.Deflate | DecompressionMethods.GZip;
```
johnsimons, line 51: Can you add Brotli support, that is the default in our Server

source/Octopus.Client/OctopusClient.cs
```diff
@@ -420,6 +420,7 @@ protected virtual OctopusResponse<TResponseResource> DispatchRequest<TResponseRe
             webRequest.Method = request.Method;
             webRequest.Headers[ApiConstants.ApiKeyHttpHeaderName] = serverEndpoint.ApiKey;
             webRequest.UserAgent = octopusCustomHeaders.UserAgent;
+            webRequest.AutomaticDecompression = DecompressionMethods.Deflate | DecompressionMethods.GZip;
```
johnsimons, line 423: Can you add Brotli support, that is the default in our Server

johnsimons (approved):
We need to support Brotli, given that is the default compression on Server now that we have gone ASP.NET core pipeline

adam-mccoy:
Adding support for Brotli would be great, but the .NET versions we're targeting don't have built-in support. We'd either have to re-target or take a dependency on a third-party implementation, which blows the scope of this item.

PR #542, opened 2020-09-22
Remove GitRefType on Release resource

ankithkonda:
Removing the GitRefType on the release resource. 

If the consumer wants to release from a tag called 1.0.0, but they also have a branch called 1.0.0, we expect a full name - refs/tags/1.0.0. Otherwise we use the branch 1.0.0

we will always assume a GitRef value is a branch, if a full ref hasn't been provided

main - branch
1.0.0 - branch
refs/heads/main - branch
refs/heads/1.0.0 - branch
refs/tags/main - tag
refs/tags/refs/heads/main - tag
refs/tags/1.0.0 - tag


tothegills (approved):
Lets see how this feels, I think there will be less friction. It's looking more like the GitHub API:

image

PR #546, opened 2020-09-28
Allow `Delete` calls to include a resource.

andrew-moore-octo:
This PR adds a way to call `Delete` with a resource as the body of the request. This is required for config-as-code runbooks, where we want to allow a user to supply a commit message (via the request body) when deleting a runbook.

Reference: https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/DELETE

### DIFF
--- a/source/Octopus.Client/IOctopusClient.cs
+++ b/source/Octopus.Client/IOctopusClient.cs
@@ -328,7 +328,7 @@ namespace Octopus.Client
         /// <param name="resource">The body of the request.</param>
         /// <param name="pathParameters">If the <c>path</c> is a URI template, parameters to use for substitution.</param>
         /// <returns>A task resource that provides details about the background task that deletes the specified resource.</returns>
-        void Delete<TResource>(string path, TResource resource, object pathParameters = null);
+        void DeleteWithResource<TResource>(string path, TResource resource, object pathParameters = null);
 
         /// <summary>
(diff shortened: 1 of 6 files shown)


source/Octopus.Client/IOctopusClient.cs
```diff
@@ -308,8 +308,9 @@ public interface IOctopusClient : IDisposable
         /// <exception cref="OctopusResourceNotFoundException">HTTP 404: If the specified resource does not exist on the server.</exception>
         /// <param name="path">The path to the resource to delete.</param>
         /// <param name="pathParameters">If the <c>path</c> is a URI template, parameters to use for substitution.</param>
+        /// <param name="resource">An optional resource to pass as the body of the request.</param>
         /// <returns>A task resource that provides details about the background task that deletes the specified resource.</returns>
-        void Delete(string path, object pathParameters = null);
+        void Delete(string path, object pathParameters = null, object resource = null);
```
TomPeters, line 313: You should also make this change to `IOctopusAsyncClient`

andrew-moore-octo: Done.

TomPeters (approved):
Looks right to me

andrew-moore-octo:
[Octopus.Client changes to support VCS runbooks](https://trello.com/c/yVzwTtqW/116-octopusclient-changes-to-support-vcs-runbooks)

PR #549, opened 2020-10-10
Add cancellation token parameter to async methods

### LINKED ISSUE #537: Add CancellationToken support on async calls
It is considered a best practice to provide a `CancellationToken` to allow for the cancellation of tasks.

It doesn't appear that any of the async code in this library has any support for passing in a CancellationToken.  This means if a query end up being long running there is no way for the caller to abort a request.  This can cause problems for things like Windows Services where the expectation is to respond quickly to shutdown requests.

yj7o5:
* Adds an optional `CancellationToken` to all the async public methods
* Update the approvals files for netcore and netframework 
* Fix breaking test due to additional default parameter at certain places 

closes #537

### DIFF
--- a/source/Octopus.Client.Tests/Operations/RegisterMachineOperationFixture.cs
+++ b/source/Octopus.Client.Tests/Operations/RegisterMachineOperationFixture.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Linq;
+using System.Threading;
 using System.Threading.Tasks;
 
 using FluentAssertions;
@@ -109,7 +110,7 @@ namespace Octopus.Client.Tests.Operations
             await client.Received().Create("/api/machines", Arg.Is<MachineResource>(m =>
                 m.Name == "Mymachine"
                 && ((ListeningTentacleEndpointResource)m.Endpoint).Uri == "https://mymachine.test.com:10930/"
-                && m.EnvironmentIds.First() == "environments-2"))
+                && m.EnvironmentIds.First() == "environments-2"), Arg.Any<object>(), Arg.Is(default(CancellationToken)))
                 .ConfigureAwait(false);
         }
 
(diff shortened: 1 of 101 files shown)


yj7o5:
I think for some of those methods that take parameterized array such as:
https://github.com/OctopusDeploy/OctopusClients/pull/549/files#diff-b4563a4a2fe786ba2dd3fa2ccdaef341R19

we can probably leave the old definition as is i.e.:
`Task<List<TAccount>> GetOfType<TAccount>(params string[] ids)`

and add another one that takes the token (not optional but required)`
`Task<List<TAccount>> GetOfType<TAccount>(CancellationToken token, params string[] ids)`

while remove this signature that I modified it to earlier:
`Task<List<TAccount>> GetOfType<TAccount>(CancellationToken token = default, params string[] ids)`

This way it doesn't break other codebases. Although its best practice to take token as the last default parameter but not sure how best to get away with this.

yj7o5:
@matt-richardson any thoughts on above? and when can this be merged and available for usage once ready. Appreciate any feedback!

matt-richardson:
> @matt-richardson any thoughts on above? and when can this be merged and available for usage once ready. Appreciate any feedback!

Hi @yj7o5 - thanks for the PR. I've passed it onto the right team here to look at. Hopefully they should get to it soon.

As for the param ordering, we could either put the cancellation token first, or add an overload that drops the `params` part, so it takes an array and a token. Right now I'm not entirely sure... 🤔 Do what you think best and the team will look at it in more detail soon.

mcasperson:
Unfortunately the QRF team did not have capacity to address this PR. We'll keep the PR open, but unfortunately we can not provide any guidance as to when it will be resolved.

yj7o5:
That's fine. Appreciate the response.

PR #551, opened 2020-10-28
Add Support for API Key Hints

andyinaus:
Create asymmetric ApiKey resources to match the changes in the Server.
Add support for the Hint property within SensitiveValue.

Used in [Server PR](https://github.com/OctopusDeploy/OctopusDeploy/pull/7542)

[Trello Card](https://trello.com/c/iJGDDeQk/398-display-masked-api-key)

### DIFF
--- a/source/Octopus.Client/Model/PropertyValueResource.cs
+++ b/source/Octopus.Client/Model/PropertyValueResource.cs
@@ -75,7 +75,8 @@ namespace Octopus.Client.Model
                                 new SensitiveValue
                                 {
                                     HasValue = jo.GetValue("HasValue")?.ToObject<bool>() ?? false,
-                                    NewValue = jo.GetValue("NewValue").ToObject<string>()
+                                    NewValue = jo.GetValue("NewValue").ToObject<string>(),
+                                    Hint = jo.GetValue("Hint")?.ToObject<string>()
                                 });
                         }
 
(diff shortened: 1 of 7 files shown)


slewis74 (commented):
What's the purpose of adding this? We have the Purpose today, so what value is a Name adding? Is there a corresponding server PR related to this that could be linked for context?

andyinaus:
> What's the purpose of adding this? We have the Purpose today, so what value is a Name adding? Is there a corresponding server PR related to this that could be linked for context?

Linked :)

adam-mccoy:
[Display masked API key](https://trello.com/c/iJGDDeQk/398-display-masked-api-key)

source/Octopus.Client/Model/PropertyValueResource.cs
```diff
@@ -75,7 +75,8 @@ public override object ReadJson(JsonReader reader, Type objectType, object exist
                                 new SensitiveValue
                                 {
                                     HasValue = jo.GetValue("HasValue")?.ToObject<bool>() ?? false,
-                                    NewValue = jo.GetValue("NewValue").ToObject<string>()
+                                    NewValue = jo.GetValue("NewValue").ToObject<string>(),
```
slewis74, line 78: I'm surprised this wasn't jo.GetValue("NewValue")**?**.ToObject<string>(), given the following line has that. Or doesn't the following not need it either?

adam-mccoy: We need Elvis for `Hint` because the property won't be there for older Servers. I could make it the same for `NewValue` but I was inclined to leave it alone.

PR #552, opened 2020-11-03
Add expiry date to API keys

adam-mccoy:
# Background

This PR is part of the [API Key Management](https://docs.google.com/document/d/1BwF_RjvOAX0Rc-6JEr3D1w88BJcyqxXuhYr8GPpV5Gc/edit#) pitch.

Allows setting and retrieving an expiry date for an API key.

Related to Server PR https://github.com/OctopusDeploy/OctopusDeploy/pull/7575.

### DIFF
--- a/source/Octopus.Client/Repositories/Async/UserRepository.cs
+++ b/source/Octopus.Client/Repositories/Async/UserRepository.cs
@@ -21,7 +21,7 @@ namespace Octopus.Client.Repositories.Async
         Task SignOut();
         Task<UserResource> GetCurrent();
         Task<SpaceResource[]> GetSpaces(UserResource user);
-        Task<ApiKeyCreatedResource> CreateApiKey(UserResource user, string purpose = null, DateTimeOffset? expiry = null);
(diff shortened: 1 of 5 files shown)


adam-mccoy:
[Set an expiration date when creating a key](https://trello.com/c/KPr0X3hZ/400-set-an-expiration-date-when-creating-a-key)

andyinaus (approved):
LGTM

source/Octopus.Client/Repositories/Async/UserRepository.cs
```diff
@@ -21,7 +21,7 @@ public interface IUserRepository :
         Task SignOut();
         Task<UserResource> GetCurrent();
         Task<SpaceResource[]> GetSpaces(UserResource user);
-        Task<ApiKeyCreatedResource> CreateApiKey(UserResource user, string purpose = null);
+        Task<ApiKeyCreatedResource> CreateApiKey(UserResource user, string purpose = null, DateTimeOffset? expires = null);
```
slewis74, line 24: Does null here mean it never expires? If so, it might be worth an Xml doc to point that out explicitly

PR #554, opened 2020-11-11
Add command and response DTOs for converting projects to version-controlled

uglybugger:
We have a need in Server to pass commit messages and other paraphernalia through when performing actions.

This PR adds a command DTO and corresponding method on the `IProjectBetaRepository` to allow us to convert projects to version-controlled and pass through a commit message.

### DIFF
--- a/source/Octopus.Client.Tests/PublicSurfaceAreaFixture.cs
+++ b/source/Octopus.Client.Tests/PublicSurfaceAreaFixture.cs
@@ -28,9 +28,9 @@ namespace Octopus.Client.Tests
             var framework = string.Concat(RuntimeInformation.FrameworkDescription.Split(' ').Take(2));
             try
             {
-                var recieved = string.Join("\r\n", lines);
+                var received = string.Join("\r\n", lines);
                 this.Assent(
-                    recieved,
+                    received,
                     new Configuration().UsingNamer(new PostfixNamer(framework))
                 );
             }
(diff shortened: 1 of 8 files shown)


source/Octopus.Client/Model/VersionControl/ConvertProjectToVersionControlledCommand.cs
```diff
@@ -0,0 +1,13 @@
+﻿using System.ComponentModel.DataAnnotations;
+
+namespace Octopus.Client.Model.VersionControl
+{
+    public class ConvertProjectToVersionControlledCommand
```
uglybugger, line 5: As discussed with @tothegills and @ankithkonda: this makes the intention of the API call clear - it's an action, not a resource. It's still hidden behind the `Beta()` interface to the `ProjectsRepository` so we can change it any time we like, but it feels like a good place to start.

source/Octopus.Client.Tests/PublicSurfaceAreaFixture.cs
```diff
@@ -28,9 +28,9 @@ public void ThePublicSurfaceAreaShouldNotRegress()
             var framework = string.Concat(RuntimeInformation.FrameworkDescription.Split(' ').Take(2));
             try
             {
-                var recieved = string.Join("\r\n", lines);
+                var received = string.Join("\r\n", lines);
```
uglybugger, line 31: Fixing seplling/tpyo.

MarkSiedle (approved):
This seems consistent with the Server work this is designed to support, LGTM 👍

PR #555, opened 2020-11-11
Expect page metadata in branch response

tothegills:
The server now returns a `ResourceCollection` of `VersionControlBranchResource` instead of array.

This relates to the server PR https://github.com/OctopusDeploy/OctopusDeploy/pull/7637


MarkSiedle (approved):
Makes sense to me and corresponds with the server PR that's in the works 👍

PR #558, opened 2020-11-27
Added BasePath to VersionControlSettingsResource

YuKitsune:
Added a `BasePath` property to the `VersionControlSettingsResource` allowing for the ability to specify a base path for the Octopus project's .ocl files as part of Config as Code.


uglybugger (approved):
LGTM :) This is already being used via a pre-release build in Server and appears to be happy.

PR #563, opened 2021-01-18
When deleting remove channel from trackedChannelBuilders

BlythMeister:
code example of issue.
Removing the default channel in a project.

```
                    var defaultChannelEditor = projectEditor.Channels.CreateOrModify("Default");
                    defaultChannelEditor.Customize(x => x.IsDefault = false);
                    defaultChannelEditor.Save();
                    projectEditor.Channels.Delete("Default");
                    projectEditor.Save();
```
the call to `projectEditor.Save();` will error as it tries to save the channel that has just been deleted.
We need to do `CreateOrModify` in order to set `IsDefault` as false to allow it to be deleted


tothegills:
Merged here: https://github.com/OctopusDeploy/OctopusClients/pull/563/files

PR #564, opened 2021-01-25
Removing unnecessary calls to GET resources after a POST/PUT

MarkSiedle:
See equivalent [Server PR](https://github.com/OctopusDeploy/OctopusDeploy/pull/8019)

Customers using an older version of Octopus should not be affected (E.g. version 2.6 that may not have returned the resource as part of a POST/PUT), as they should be using an older version of OctopusClients - [see compatibility documentation](https://octopus.com/docs/support/compatibility).


MarkSiedle:
Closing til we get server unblocked re. the library endpoints first.

PR #565, opened 2021-01-29
Separate deployment settings API and resource

emilol:
Adds support for deployment settings endpoints.

*Note:* this is a breaking change for VCS projects.

## Background

See [Config-as-Code: Project Settings API](https://github.com/OctopusDeploy/Architecture/blob/master/OctopusServer/ConfigAsCode/ConfigAsCodeSettingsAPI.md)

# Before

All projects get/update deployment-related settings via get/modify Project.

# After

Get/Modify Project endpoints:
- Maintain backwards compatibility for db projects. 
- Deployment Settings are no longer populated for VCS projects 
   _in future this will be a [Bad Request](https://github.com/OctopusDeploy/Architecture/blob/master/OctopusServer/ConfigAsCode/ConfigAsCodeSettingsAPI.md#considered-solution-3) - see [implementation](https://github.com/OctopusDeploy/OctopusDeploy/pull/8073/files#diff-4809639251bd9cfba039cb763e088a5632123ede8e8c11c9b8765e276481123cR46) (internal link)_

Get/Modify Deployment Settings endpoints:
- Should be used going forward for both db and CaC projects.

Config as Code, here we come!

### DIFF
--- a/source/Octopus.Client.Tests/PublicSurfaceAreaFixture.ThePublicSurfaceAreaShouldNotRegress..NETCore.approved.txt
+++ b/source/Octopus.Client.Tests/PublicSurfaceAreaFixture.ThePublicSurfaceAreaShouldNotRegress..NETCore.approved.txt
@@ -146,6 +146,7 @@ Octopus.Client
     Octopus.Client.Repositories.Async.IDefectsRepository Defects { get; }
     Octopus.Client.Repositories.Async.IDeploymentProcessRepository DeploymentProcesses { get; }
     Octopus.Client.Repositories.Async.IDeploymentRepository Deployments { get; }
+    Octopus.Client.Repositories.Async.IDeploymentSettingsRepository DeploymentSettings { get; }
     Octopus.Client.Repositories.Async.IEnvironmentRepository Environments { get; }
     Octopus.Client.Repositories.Async.IFeedRepository Feeds { get; }
     Octopus.Client.Repositories.Async.IInterruptionRepository Interruptions { get; }
@@ -318,6 +319,7 @@ Octopus.Client
     Octopus.Client.Repositories.Async.IDefectsRepository Defects { get; }
     Octopus.Client.Repositories.Async.IDeploymentProcessRepository DeploymentProcesses { get; }
     Octopus.Client.Repositories.Async.IDeploymentRepository Deployments { get; }
+    Octopus.Client.Repositories.Async.IDeploymentSettingsRepository DeploymentSettings { get; }
     Octopus.Client.Repositories.Async.IEnvironmentRepository Environments { get; }
     Octopus.Client.Repositories.Async.IEventRepository Events { get; }
     Octopus.Client.Repositories.Async.IFeaturesConfigurationRepository FeaturesConfiguration { get; }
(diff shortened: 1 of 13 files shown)


source/Octopus.Client/Model/ProjectResource.cs
```diff
@@ -86,21 +86,6 @@ public ProjectResource(string id, string name, string slug) : this()
         [Writeable]
         public List<string> IncludedLibraryVariableSetIds { get; set; }
```
emilol, line 88: I know we can't _really_ remove these, but a girl can dream, can't she?
What's the right approach here for maintaining backwards compat - obsolete tags?

source/Octopus.Client/Repositories/DeploymentSettingsRepository.cs
```diff
@@ -0,0 +1,72 @@
...
+        private readonly IOctopusRepository repository;
+        private readonly IOctopusClient client;
+
+        public DeploymentSettingsRepositoryBeta(IOctopusRepository repository)
+        {
+            this.repository = repository;
+            client = repository.Client;
+        }
+
+        public DeploymentSettingsResource Get(ProjectResource projectResource, string gitref = null)
```
emilol, line 42: API design choice: should these be based on `ProjectId` or `DeploymentSettingsId`? I defaulted to project.

source/Octopus.Client/Model/ProjectResource.cs
MarkSiedle: Yes, `Obsolete` and try and guide the consumer to the new way of doing things. I.e. Something like:

```
[Obsolete("Use `DeploymentSettingsResource` instead on the `deploymentsettings` API.")]
```

source/Octopus.Client.Tests/PublicSurfaceAreaFixture.ThePublicSurfaceAreaShouldNotRegress..NETCore.approved.txt
```diff
@@ -6772,6 +6781,15 @@ Octopus.Client.Repositories
     void Paginate(String[], String[], Func<ResourceCollection<DeploymentResource>, Boolean>)
     void Paginate(String[], String[], String[], Func<ResourceCollection<DeploymentResource>, Boolean>)
   }
+  interface IDeploymentSettingsRepository
+  {
+    Octopus.Client.Repositories.IDeploymentSettingsRepositoryBeta Beta()
```
MarkSiedle, line 6786: Slight naming inconsistency, these are currently `ThingBetaRepository`

```suggestion
    Octopus.Client.Repositories.IDeploymentSettingsBetaRepository Beta()
```

source/Octopus.Client/Repositories/DeploymentSettingsRepository.cs
MarkSiedle: This looks correct. We should follow the existing pattern used by deployment process, since this model aligns nicely with that. We just grab the appropriate link from the project then 👍

MarkSiedle (commented):
Looks good, couple of thoughts below on obsolete etc.

MarkSiedle (approved):
This looks consistent with the current approach to me, and all previous feedback has been actioned nicely, thank you. LGTM 👍

PR #567, opened 2021-02-03
Version Control Settings and Credentials refactor

YuKitsune:
This PR refactors the way Version Control settings are structured within the Project Resource.

# Before
`VersionControlSettings` was a property on the `ProjectResource` which contained everything required to configure and maintain a version controlled project. If a project was not version controlled, then all properties in the `VersionControlSettings` would be `null`.

# After
`VersionControlSettings` has been replaced with `PersistenceSettings`. `PersistenceSettings` can either be an instance of `DatabasePersistenceSettings` (for non-version controlled projects), or `VersionControlSettings` (for version controlled projects).

The `Username` and `Password` properties have also been removed from the `VersionControlSettings`, and replaced with a `VersionControlCredentials` class, which can either be `UsernamePasswordVersionControlCredentials` (when using username and password/PAT) or `AnonymousVersionControlCredentials` (when authentication is not required).

This reduces the overall amount of `null`'s in the code, and reduces the amount of code for checking whether or not a project is version controlled, and what type of authentication is being used.


emilol (approved):
This all seems sensible to me based on the conversations we've had 👍 

Have you had a chance to reference the change from Server and see if it plays nicely?

YuKitsune:
Hey @emilol 
Working on that at the moment in this PR: https://github.com/OctopusDeploy/OctopusDeploy/pull/8076

YuKitsune:
> Have you had a chance to reference the change from Server and see if it plays nicely?
Sure have, it's working well with Servers changes

gupta-kartik (approved):
Changes LGTM

PR #569, opened 2021-03-01
Added Architecture property to the MachineBaseResource class to allow…

MissedTheMark:
… querying of same through the API.


tothegills (approved):
First one!! Woohoo! 👍

PR #573, opened 2021-03-10
Add non-beta path for Deployment Settings endpoints

emilol:
# Background
Deployment Settings props on `ProjectResource` have been [deprecated](https://github.com/OctopusDeploy/OctopusClients/blob/8a6acb14a9e412d5829560cebd2963d86a89dccd/source/Octopus.Client/Model/ProjectResource.cs#L94). This was done to unlock functionality for Config as Code - allowing these settings to persist to git. 

The supported path forward for Database Projects (i.e. traditional non-version controlled projects in Octopus Deploy) is to modify these settings via 

```http
POST api/projects/{projectId}/deploymentsettings
``` 

In future `POST projects/{projectId}/{gitRef}/deploymentsettings` will also be supported.

## Current workaround
```csharp
octopusRepository.Client.Post($"projects/{projectResource.Id}/deploymentsettings", deploymentSettingsResource);
```

# Results

This PR provides a strong typing for the supported (i.e. non-beta) endpoint for DeploymentSettings for DB projects.

```csharp
octopusRepository.DeploymentSettings.Modify(projectResource, deploymentSettingsResource);
```

### DIFF
--- a/source/Octopus.Client/Repositories/Async/DeploymentSettingsRepository.cs
+++ b/source/Octopus.Client/Repositories/Async/DeploymentSettingsRepository.cs
@@ -12,7 +12,7 @@ namespace Octopus.Client.Repositories.Async
         Task<DeploymentSettingsResource> Modify(ProjectResource project, DeploymentSettingsResource deploymentSettings);
     }
 
-    class DeploymentSettingsRepository : IDeploymentSettingsRepository
+    internal class DeploymentSettingsRepository : IDeploymentSettingsRepository
     {
         private readonly IDeploymentSettingsBetaRepository beta;
         private readonly IOctopusAsyncClient client;
@@ -22,6 +22,7 @@ namespace Octopus.Client.Repositories.Async
             beta = new DeploymentSettingsBetaRepository(repository);
             client = repository.Client;
         }
+
         public IDeploymentSettingsBetaRepository Beta()
         {
             return beta;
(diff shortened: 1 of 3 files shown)


source/Octopus.Client/Repositories/Async/DeploymentSettingsRepository.cs
```diff
@@ -6,68 +7,92 @@ namespace Octopus.Client.Repositories.Async
...
-            if (!string.IsNullOrWhiteSpace(gitref))
-            {
-                var branchResource = await repository.Projects.Beta().GetVersionControlledBranch(projectResource, gitref);
+            if (!(project.PersistenceSettings is VersionControlSettingsResource settings))
+                return await repository.DeploymentSettings.Get(project);
 
-                return await client.Get<DeploymentSettingsResource>(branchResource.Link("DeploymentSettings"));
-            }
+            gitref = gitref ?? settings.DefaultBranch;
+            var branch = await repository.Projects.Beta().GetVersionControlledBranch(project, gitref);
```
emilol, line 78: the change here to lookup the branch even when it's default was deliberate - previously it was relying on the project link, which coincidentally we've fully resolving to the default at the moment. 

We probably shouldn't do that either. At a project level we should be passing a partially resolved link.

MarkSiedle (approved):
Looks sensible to me 👍

PR #574, opened 2021-03-22
Add IGetAll<> interface to the IChannelRepository

zentron:
We support the `/all` endpoint in[ OctopusServer for Channels](https://github.com/OctopusDeploy/OctopusDeploy/blob/1f1a7a667daa86a14e8dee4f448d346c3b673389/source/Octopus.Server/Web/Api/Actions/Channels/ChannelsNancyModule.cs#L27) so in to synchronize the surface area we should probably ensure the C# client can interact with it.

### DIFF
--- a/build/Build.cs
+++ b/build/Build.cs
@@ -177,7 +177,7 @@ class Build : NukeBuild
             .SetFiles(files.Select(x => x.ToString()).ToArray())
             .SetProcessToolPath(RootDirectory / "certificates" / "signtool.exe")
             .SetTimestampServerDigestAlgorithm("sha256")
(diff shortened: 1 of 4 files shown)


build/Build.cs
```diff
@@ -170,14 +181,31 @@ private void SignBinaries(AbsolutePath path)
     {
         Info($"Signing binaries in {path}");
         var files = path.GlobDirectories("**").SelectMany(x => x.GlobFiles("Octopus.*.dll"));
+
```
zentron, line 184: These unrelated changes were required to have the build complete successfully.

uglybugger (approved):
LGTM :)

PR #576, opened 2021-04-07
Expose new RetentionDefaultConfigurationResource

laurencee:
Client resource for this [Server PR](https://github.com/OctopusDeploy/OctopusDeploy/pull/8589)

### DIFF
--- a/source/Octopus.Client/Model/RetentionDefaultConfigurationResource.cs
+++ b/source/Octopus.Client/Model/RetentionDefaultConfigurationResource.cs
@@ -4,11 +4,6 @@ namespace Octopus.Client.Model
 {
     public class RetentionDefaultConfigurationResource : Resource
     {
-        public RetentionDefaultConfigurationResource()
-        {
-            Id = "retention-default";
-        }
-
         [Writeable]
         public int? RetentionDays { get; set; }
     }
--- a/source/Octopus.Client/Model/RetentionDefaultConfigurationResource.cs
+++ b/source/Octopus.Client/Model/RetentionDefaultConfigurationResource.cs
@@ -1,0 +1,10 @@
+﻿using Octopus.Client.Extensibility.Attributes;
+
+namespace Octopus.Client.Model
+{
+    public class RetentionDefaultConfigurationResource : Resource
+    {
+        [Writeable]
+        public int? RetentionDays { get; set; }
+    }
+}
--- a/source/Octopus.Client/Model/RetentionDefaultConfigurationResource.cs
+++ b/source/Octopus.Client/Model/RetentionDefaultConfigurationResource.cs
@@ -4,6 +4,11 @@ namespace Octopus.Client.Model
 {
     public class RetentionDefaultConfigurationResource : Resource
     {
+        public RetentionDefaultConfigurationResource()
+        {
+            Id = "retention-default";
+        }
+
         [Writeable]
         public int? RetentionDays { get; set; }
     }
--- a/source/Octopus.Client.Tests/PublicSurfaceAreaFixture.ThePublicSurfaceAreaShouldNotRegress..NETCore.approved.txt
+++ b/source/Octopus.Client.Tests/PublicSurfaceAreaFixture.ThePublicSurfaceAreaShouldNotRegress..NETCore.approved.txt
@@ -4295,6 +4295,14 @@ Octopus.Client.Model
   {
     static System.String Name
   }
+  class RetentionDefaultConfigurationResource
+    Octopus.Client.Extensibility.IResource
+    Octopus.Client.Model.IAuditedResource
+    Octopus.Client.Model.Resource
+  {
+    .ctor()
+    Nullable<Int32> RetentionDays { get; set; }
+  }
   class RetentionPeriod
     IEquatable<RetentionPeriod>
   {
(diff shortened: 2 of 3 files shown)


pawelpabich:
Don't we need to expose a new method that takes `RetentionDefaultConfigurationResource` as a parameter on one of the repositories?

laurencee:
> Don't we need to expose a new method that takes `RetentionDefaultConfigurationResource` as a parameter on one of the repositories?

Good question, I was using other configuration resources as inspriations which don't currently have their own repository.

I'm not sure which approach is the current correct approach.

https://github.com/OctopusDeploy/OctopusClients/blob/master/source/Octopus.Client/Model/SmtpConfigurationResource.cs
https://github.com/OctopusDeploy/OctopusClients/blob/master/source/Octopus.Client/Model/ServerConfigurationResource.cs
https://github.com/OctopusDeploy/OctopusClients/blob/master/source/Octopus.Client/Model/MaintenanceConfigurationResource.cs

They were being used via the generic repository in these places:

https://github.com/OctopusDeploy/OctopusDeploy/blob/master/source/Octopus.E2ETests[…].API/Scripts/ConfigurationAsCode/NamesInsteadOfIdsTestScript.cs
https://github.com/OctopusDeploy/OctopusDeploy/blob/master/source/Octopus.IntegrationTests/Server/Permissions/SmtpConfigurationFixture.cs#L47
https://github.com/OctopusDeploy/OctopusDeploy/blob/master/source/Octopus.IntegrationTests/Server/Permissions/MaintenanceConfigurationFixture.cs
https://github.com/OctopusDeploy/Octofront/blob/master/source/Octofront.Core/Hosted/Octopus/MaintenanceModeConfigurator.cs#L45

laurencee:
@benPearce1 Are you able to provide input into this?

pawelpabich:
Ah, so we are using the client directly and you *simply* need to know which URL to use :). 
@slewis74 do you have an opinion on this?

source/Octopus.Client/Model/RetentionDefaultConfigurationResource.cs
```diff
@@ -0,0 +1,10 @@
+﻿using Octopus.Client.Extensibility.Attributes;
+
+namespace Octopus.Client.Model
+{
+    public class RetentionDefaultConfigurationResource : Resource
```
benPearce1, line 5: This class is going to need to set the `Id` property to a well known value. Similar to https://github.com/OctopusDeploy/Octofront/blob/f59e858ac93ed7e003e2b253029b3baaed961559/source/Octofront.Core/Hosted/Instances/DynamicWorkerConfiguration.cs#L5

benPearce1:
We already use this similar pattern with Dynamic Worker Configuration in [Octofront](https://github.com/OctopusDeploy/Octofront/blob/f59e858ac93ed7e003e2b253029b3baaed961559/source/Octofront.Core/Hosted/DynamicWorkers/DynamicWorkerServiceConfigurator.cs#L61).
As long as the class inherits from Resource and sets the id to a well known value, you can read and modify via the Configuration Repository in Client.

benPearce1 (changes requested):

benPearce1:
And the same pattern is used in our Auth extensions
https://github.com/OctopusDeploy/OpenIDConnectAuthenticationProviders/blob/9f60c5d5f4fe09e3d2e44f3057b0f05052429c4f/source/Client.OctopusID/Configuration/OctopusIDConfigurationResource.cs

source/Octopus.Client/Model/RetentionDefaultConfigurationResource.cs
pawelpabich: Why do we need to specify the id here?  We have a dedicated API endpoint for this resource: https://github.com/OctopusDeploy/OctopusDeploy/blob/le/file-retention/source/Octopus.Server/Web/Controllers/Configuration/Retention/ModifyRetentionDefaultConfigurationController.cs#L39.

I think this is not the case for DW which is an exception here.

benPearce1: To call this endpoint would require a new repository in Octopus.Client. By specifying the Id on the Resource model, you can go via the existing Configuration Repository and don't actually need the new controller in Server. 
Example: https://octopus.com/docs/administration/managing-infrastructure/server-configuration/managing-configuration#net-c
And the corresponding Client Resource https://github.com/OctopusDeploy/OctopusClients/blob/master/source/Octopus.Client/Model/WebPortalConfigurationResource.cs

pawelpabich: We need a controller because we have some business rules to enforce. It's not a simple CRUD thing.

pawelpabich: Similar to how License is treated. It lives in the Configuration table but has its own endpoints.

laurencee: Just to clarify this one as this was discussed in this mornings sync, we are talking about 2 different repositories here.

The one that requires the Id to be set is the general purpose `Configuration` repository while for this particular case we can just use the `Client` repository instead. The client repository allows us to go to a specific/custom endpoint while `Configuration` has its own endpoint.

Example of this already being done from the [MaintenanceConfigurationResource](https://github.com/OctopusDeploy/OctopusClients/blob/master/source/Octopus.Client/Model/MaintenanceConfigurationResource.cs):

https://github.com/OctopusDeploy/Octofront/blob/master/source/Octofront.Core/Hosted/Octopus/MaintenanceModeConfigurator.cs#L45

source/Octopus.Client/Model/RetentionDefaultConfigurationResource.cs
```diff
@@ -0,0 +1,10 @@
...
+    {
+        [Writeable]
+        public int? RetentionDays { get; set; }
```
benPearce1, line 8: Is a null here the same as no default?

benPearce1 (approved):
Since we are all on the same page now, LGTM

PR #581, opened 2021-05-04
Improve Tentacle command line validation

matt-richardson:
This PR changes the type of exception thrown when validation arguments for the `register-with` and `register-worker` commands from `ArgumentException` to a new custom type `InvalidRegistrationArgumentsException`, so we can handle it better.

The new type inherits from `ArgumentException`, so this should be backwards compatible.

This PR is required by https://github.com/OctopusDeploy/OctopusTentacle/pull/282.

## Review

Please confirm my thinking that this is backwards compatible.
Other than that, :eyes: and a 💚 build should be sufficient for this PR.


shaunmarx (approved):
Looks good to me 👍🏻

PR #582, opened 2021-05-06
ILMerge the netstandard version as well

matt-richardson:
We ran into an issue internally where the dependencies added in https://github.com/OctopusDeploy/OctopusClients/pull/578 broke downstream builds (https://build.octopushq.com/viewLog.html?buildId=2323447&tab=buildResultsDiv&buildTypeId=Octofront_PublicE2eTests#testNameId-7449192310281387661).

We had previously only ILMerged the net452 version, as dotnet core 2.x didn't support it. This support was added in 3.x (i think).

We had a mention in our [docs about this](https://github.com/OctopusDeploy/docs/blob/master/docs/octopus-rest-api/octopus.client/getting-started.md), but we'd forgotten.

Now we're building with netcore 5 SDK, we can ILMerge ~all the things~ again.

**EDIT**: Turns out if we merge Octopus.TinyTypes.*, we end up with issues deserializing things as we end up with the wrong types coming out the end. So, we've (@MJRichardson, @droyad, @uglybugger and I) changed it so that we only merge the libraries we use internal (newtonsoft.json & octodiff), and we use nuget dependencies to manage pull in the other dependencies.


droyad:
@uglybugger Was the intent of the TinyType addition to expose the tiny types to the consumer? If so this change will make those TinyTypes _different_ types to those defined in the `Octopus.TinyType` project. This means if the consumer uses both they'd have to map between them.

uglybugger (approved):
Sneaky! Well spotted. Thanks for fixing this one!

uglybugger:
> @uglybugger Was the intent of the TinyType addition to expose the tiny types to the consumer? If so this change will make those TinyTypes _different_ types to those defined in the `Octopus.TinyType` project. This means if the consumer uses both they'd have to map between them.

Hmm. Good question. I need to think about this a little more.

_Eventually_ we would like to use the resource types defined `Octopus.Server.MessageContracts`. That's not what _this_ change is about, but eventually we'll want that.

_Right now_, we want `Octopus.Client` to continue using its existing DTOs (which don't use tiny types) so this shouldn't change any behaviours as far as external consumers are concerned. However, our own integration tests use some message contracts whih are defined in the `OctopusDeploy` solution and not exposed at all via `Octopus.Client`. These are the ones which cause the problem, because they use tiny types.

I suspect you're right, in that it might cause our integration tests to fail to compile/link correctly. It might also cause the entire solution to fail to compile depending on where the message contracts are defined (which is in the process of moving).

Could we potentially hold off on merging this change until perhaps early next week? I'm happy to test this PR's NuGet package and see what the implications for Server might be, but I just don't have the capacity today.

matt-richardson:
> Could we potentially hold off on merging this change until perhaps early next week? I'm happy to test this PR's NuGet package and see what the implications for Server might be, but I just don't have the capacity today.

Sure. I was trying to get this through to unblock https://github.com/OctopusDeploy/OctopusTentacle/pull/282, but there's no urgency on that one. I could either push it through with the extra deps, and fix it up later, or just sit on it for a few days.

build/Build.cs
```diff
@@ -95,30 +95,30 @@ class Build : NukeBuild
...
-            new ILRepack(repackSettings).Repack();
-
-            DeleteDirectory(inputFolder);
-            MoveDirectory(outputFolder, inputFolder);
-    });
+                var inputFolder = OctopusClientFolder / "bin" / Configuration / target;
+                var outputFolder = OctopusClientFolder / "bin" / Configuration / $"{target}Merged";
+                EnsureExistingDirectory(outputFolder);
+
+                var assemblyPaths = inputFolder.GlobFiles("NewtonSoft.Json.dll", "Octodiff.*", "Octopus.TinyTypes.dll", "Octopus.TinyTypes.Json.dll", "Octopus.TinyTypes.TypeConverters.dll");
```
matt-richardson, line 104: "Octopus.TinyTypes.dll", "Octopus.TinyTypes.Json.dll", "Octopus.TinyTypes.TypeConverters.dll" shouldn't be added here.
They need to be in the nuspec though.

source/Octopus.Client.E2ETests/InlinedDependenciesFixture.cs
```diff
@@ -47,43 +47,19 @@ public void OneTimeTearDown()
...
-                .FirstOrDefault(x => x.FullName == "Newtonsoft.Json.JsonConvert")
-                .Should()
-                .BeNull("We aren't able to ilmerge Newtonsoft.Json in yet as it's only available from netcore 3");
-        }
 
         [Test]
-        public void HasNotInlinedOctodiff()
+        [TestCase("Newtonsoft.Json", "Newtonsoft.Json.JsonConvert")]
+        [TestCase("Octodiff", "Octodiff.Core.DeltaBuilder")]
+        [TestCase("Octopus.TinyTypes", "Octopus.TinyTypes.CaseInsensitiveStringTinyType")]
```
matt-richardson, line 54: no longer true

build/Build.cs
matt-richardson:
```suggestion
                var assemblyPaths = inputFolder.GlobFiles("NewtonSoft.Json.dll", "Octodiff.*");
```

source/Octopus.Client.E2ETests/InlinedDependenciesFixture.cs
```diff
@@ -47,43 +47,19 @@ public void OneTimeTearDown()
...
+        [TestCase("Octopus.TinyTypes.Json", "TinyTypes.Json.TinyTypeJsonConverter")]
+        [TestCase("Octopus.TinyTypes.TypeConverters", "Octopus.TinyTypes.TypeConverters.TinyTypeConverters")]
```
matt-richardson, line 56:
```suggestion
```

uglybugger (approved):
As discussed - this looks sensible to me. Thank you again for making the changes!

.nuke/build.schema.json
```diff
@@ -68,7 +68,6 @@
               "Clean",
               "Compile",
               "CopyToLocalPackages",
-              "Default",
```
matt-richardson, line 71: We removed the Default target recently, but never updated this file

build/Build.cs
```diff
@@ -25,7 +25,7 @@ class Build : NukeBuild
     // ARGUMENTS
     //////////////////////////////////////////////////////////////////////
     [Parameter] readonly string Configuration = "Release";
-    [Parameter] readonly string SigningCertificatePath = "./certificates/OctopusDevelopment.pfx";
+    [Parameter] readonly string SigningCertificatePath = RootDirectory / "certificates" / "OctopusDevelopment.pfx";
```
matt-richardson, line 28: Crashed locally for some reason as it was a relative path

build/Build.cs
```diff
@@ -95,58 +95,67 @@ class Build : NukeBuild
...
-            new ILRepack(repackSettings).Repack();
-
-            DeleteDirectory(inputFolder);
-            MoveDirectory(outputFolder, inputFolder);
-    });
+                var inputFolder = OctopusClientFolder / "bin" / Configuration / target;
+                var outputFolder = OctopusClientFolder / "bin" / Configuration / $"{target}Merged";
+                EnsureExistingDirectory(outputFolder);
+
+                var assemblyPaths = inputFolder.GlobFiles("NewtonSoft.Json.dll", "Octodiff.*");
```
matt-richardson, line 104: exe for net452, dll for netframework

build/Build.cs
```diff
@@ -95,58 +95,67 @@ class Build : NukeBuild
...
         try
         {
-            ReplaceTextInFiles(OctopusClientFolder / "Octopus.Client.nuspec", "<version>$version$</version>",
+            ReplaceTextInFiles(octopusClientNuspec, "<version>$version$</version>",
                 $"<version>{OctoVersionInfo.FullSemVer}</version>");
 
-            DotNetPack(_ => _
-            .SetProject(OctopusClientFolder)
-            .SetProcessArgumentConfigurator(args =>
+            //ensure our dependencies here match the versions in the csproj
```
matt-richardson, line 134: Bit hacky... Not sure if there's a better way though.

build/Build.cs
uglybugger, line 28: Ooh! Shiny Path.Combine operator from NUKE. I didn't know about this :)

matt-richardson: Yeah, tis very nice.

matt-richardson:
While this isn't perfect, I think it's better than where we are right now.

Unfortunately, while this fixes some of the stuff, we're now back in the realms of `newtonsoft.json` binding issues.
`Tentacle` -> `OctopusClients` -> `Octopus.TinyTypes` -> `newtonsoft.json`

Merging now - we can tackle the `newtonsoft.json` as a separate pr.

PR #586, opened 2021-05-27
Use self link to get a resource after creating/modifying

YuKitsune:
# Background
We've found that some `POST`/`PUT` endpoints aren't guaranteed to return the same information you'd find in a `GET` request to the same resource. This means we require another round trip to the server to make sure we've got a trustworthy resource to work with.

# Results
This PR changes the behaviour of the `Create` and `Update` methods on the `OctopusClient` and `OctopusAsuntClient` to send a new `GET` request using the returned resources `Self` link.
This solution isn't ideal, but it'll keep us moving forward until a proper fix is in place.

Once a proper fix is in place for this in Octopus Server, the changes from this PR can be reverted.

# Testing
None of the existing tests are really kitted out for testing the changed made in this PR, and considering this is a temporary solution, I don't see the value in having a new unit or integration test.
I think so long as the existing tests in projects that depend on `Octopus.Client` are passing, then we can be confident that we haven't broken anything.


source/Octopus.Client/UrlHelper.cs
```diff
@@ -0,0 +1,21 @@
+using Octopus.Client.Extensibility;
+
+namespace Octopus.Client
+{
+    internal static class UrlHelper
```
MarkSiedle, line 5: A year ago I would have done it this way as well, but we've been burned by code that starts this way and are trying to limit statics to extension methods where possible.

Would there be value in making a `ResourceSelfLinkExtractor` class, who's sole job is to extract a link from a given resource? Then we just new that up as a property on the `OctopusClient` and use it multiple times.

E.g. pseudocode:
```
class OctopusClient {
   readonly ResourceSelfLinkExtractor resourceSelfLinkExtractor =  new ResourceSelfLinkExtractor();
   ...
   // Then inside whatever methods we need...
   var selfLink = resourceSelfLinkExtractor.GetSelfUrlOrNull(response.ResponseResource) ?? path;
   ...
}
```

It seems small, but we may get a lot more flexibility/control in the future if we ever wanted to tweak that during tests, and this stops static helper classes/methods like this spreading via copy/pasta.

YuKitsune: Looking back on it, I can see how this is much more maintainable. I've updated it as you've suggested in ab6490704779d9578decc92ecaa8f3e46a27407c

MarkSiedle (approved):
Looks sensible to me 👍 Ty

PR #588, opened 2021-06-04
Added VCS compatibility to ChannelRepository

YuKitsune:
# Background

As part of Config-as-Code, new endpoints for channels have been made available which allow for a project ID and git ref to be specified.

# Results

The `ChannelRepository.Create` method now checks whether or not the target project is version controlled. If so, then the `projects/{projectId}/{gitRef}/channels` endpoint is used to create the channel, otherwise, the `channels` endpoint will be used. This provides backwards compatibility for existing implementations.

A new `ICreateProjectScoped<TResource>` interface has been added which allows for project-scoped resources to be created (E.g. `channels.Create(projectResource, channelResource);`) (Project ID is still required in the `ChannelResource`)

The `Modify` and `Delete` methods will continue to function without any changes, as Octopus Server adjusts the `Self` link for version controlled documents.

However, the `Get` method presents a challenge since there is only an `idOrHref` parameter, and no way to know which project or gitRef to search in. I think this should to be addressed separately.


MarkSiedle (commented):
Since we already have a ChannelEditor with a `CreateOrModify`, is it worth centralising these changes in there, or are there reasons for introducing a separate Create method here?

YuKitsune:
> Since we already have a ChannelEditor with a `CreateOrModify`, is it worth centralising these changes in there, or are there reasons for introducing a separate Create method here?

My understanding was the the `ChannelEditor` took a dependency on `ChannelRepository`, and used that for it's CRUD operations, so these changes would cover the `CreateOrModify` methods

MarkSiedle:
Ah very nice. That makes sense now. Thank you.

MarkSiedle (dismissed):

source/Octopus.Client/Repositories/Async/ProjectRepository.cs
```diff
@@ -164,5 +167,29 @@ public Task<VersionControlBranchResource> GetVersionControlledBranch(ProjectReso
...
+
+        public async Task<ChannelResource> GetChannel(ProjectResource projectResource, string gitRef, string idOrName)
+        {
+            projectResource.EnsureVersionControlled();
+
+            VersionControlBranchResource branch = await GetVersionControlledBranch(projectResource, gitRef);
+            var url = $"{branch.Link("Channels")}/{idOrName}";
+
+            return await client.Get<ChannelResource>(url);
+        }
```
YuKitsune, line 193: Each of these will send two requests. One to get the list of branches, and one to get the channel from the URL specified in the `VersionControlBranchResource`. Ideally, we could just sub the provided `gitRef` into the URL, though it doesn't look like we have that kind of capability (Unless I'm just blind, in which case, please let me know)

I've left these under the Beta repository so we can come back and change it later without too much fuss.

Same goes for the non-async repository.

YuKitsune:
I've got a separate PR up which will make adding version control compatibility to Channels much easier: https://github.com/OctopusDeploy/OctopusClients/pull/589

source/Octopus.Server.Client/Repositories/Async/DeploymentSettingsRepository.cs
```diff
@@ -44,7 +44,7 @@ public async Task<DeploymentSettingsResource> Get(ProjectResource project)
                 throw new NotSupportedException(
                     $"Version Controlled projects are still in Beta. Use {nameof(IDeploymentSettingsBetaRepository)}.");
 
-            await client.Put(deploymentSettings.Link("Self"));
```
MarkSiedle, line 47: Good catch. I'm guessing this never worked.

MarkSiedle (approved):
Looks reasonable to me. We're giving the user a hand if they don't realise they should provide a project argument for VCS projects 👍

emilol (approved):
This is looking great! 

We may find once we get this in server that there are further changes we need to make, but since the changes are additive it would be good to get this merged as-is and iterate later 👍

YuKitsune:
Thanks! The related Server PR seems fairly happy with these changes. Will get this merged.

PR #589, opened 2021-06-08
Added a Repository implementation for project-scoped resources

YuKitsune:
# Background
As we move towards a world where where project scoped resources (E.g. Channels) are accessed using the project route (`/projects/{projectId}/channels`) rather than their standalone routes (`/channels`), we're gonna see our C# client library change to reflect that.

# Results
This PR introduces a new Repository implementation based off of the `BasicRepository<TResource>`, which has project scoped methods for creating and getting resources.

Modifying and Deleting resources _shouldn't_ need their own project-scoped methods, as they use the "Self" links populated by Octopus Server.

This _should_ make converting new resources to use the project-scoped routes much easier in the future. the only changes that will be required are:
1. Ensure the resource implements the `IHaveProject` interface
2. Change the repositories base class from `BasicRepository` to `ProjectScopedRepository`

### DIFF
--- a/source/Octopus.Server.Client/Repositories/ProjectScopedRepository.cs
+++ b/source/Octopus.Server.Client/Repositories/ProjectScopedRepository.cs
@@ -41,44 +41,14 @@ namespace Octopus.Client.Repositories
             return Client.Create(link, resource, pathParameters);
         }
 
-        public TResource Get(ProjectResource projectResource, string idOrHref)
+        public TResource Get(ProjectResource projectResource, string id)
         {
             ThrowIfServerVersionIsNotCompatible();
 
-            if (string.IsNullOrWhiteSpace(idOrHref))
+            if (string.IsNullOrWhiteSpace(id))
                 return null;
-            var link = projectResource.Link(CollectionLinkName);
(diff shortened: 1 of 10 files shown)


source/Octopus.Server.Client/Repositories/ProjectScopedRepository.cs
```diff
@@ -0,0 +1,58 @@
...
+            return Client.Create(link, resource, pathParameters);
+        }
+
+        public TResource Get(ProjectResource projectResource, string id)
+        {
+            ThrowIfServerVersionIsNotCompatible();
+
+            if (string.IsNullOrWhiteSpace(id))
+                return null;
+            var link = $"{projectResource.Link(CollectionLinkName)}/{id}";
```
YuKitsune, line 50: This felt very wrong, but it works. Wondering if there is a nicer way to go about this.
Same for the async variant

uglybugger (approved):
This looks sensible to me.

As much as I don't like doubling down on code we'd like to make _less_ clever rather than more, we also need to pick our battles. This seems to be a sensible step to take in the interim before we strip lots of this logic out of the client entirely - but that's a whole different kettle of cephalopods.

PR #591, opened 2021-06-09
Remove deprecation message when using Tentacle.exe

desruc:
# Background

A deprecation message is shown when Octopus Client is used from the server/tentacle install location; however, when running legitimate `Tentacle.exe` commands, you still get the message.

# Results

This PR adds an optional `processName` parameter to the `CheckBasedOnApplication` function. If `processName` matches the active process, it returns out of the function and doesn't show the message.

The Tentacle check has been updated.


source/Octopus.Server.Client/LocationChecker.cs
```diff
@@ -18,13 +16,14 @@ public static void CheckAssemblyLocation()
...
                     return;
-                CheckBasedOnApplication(octopusRegNode, "Server");
-                CheckBasedOnApplication(octopusRegNode, "Tentacle");
+                // CheckBasedOnApplication(octopusRegNode, "Server");
+                CheckBasedOnApplication(octopusRegNode, "Tentacle", "tentacle");
             }
         }
 
-        private static void CheckBasedOnApplication(RegistryKey octopusRegNode, string application)
+        private static void CheckBasedOnApplication(RegistryKey octopusRegNode, string application, string processName = "")
```
droyad, line 24: No need to make this optional, all code paths should supply a value

source/Octopus.Server.Client/LocationChecker.cs
```diff
@@ -18,13 +16,14 @@ public static void CheckAssemblyLocation()
             {
                 if (octopusRegNode == null) 
                     return;
-                CheckBasedOnApplication(octopusRegNode, "Server");
-                CheckBasedOnApplication(octopusRegNode, "Tentacle");
+                // CheckBasedOnApplication(octopusRegNode, "Server");
```
droyad, line 19: We should remove this commented out code
```suggestion
```

source/Octopus.Server.Client/LocationChecker.cs
```diff
@@ -18,13 +16,14 @@ public static void CheckAssemblyLocation()
...
+                CheckBasedOnApplication(octopusRegNode, "Tentacle", "tentacle");
             }
         }
 
-        private static void CheckBasedOnApplication(RegistryKey octopusRegNode, string application)
+        private static void CheckBasedOnApplication(RegistryKey octopusRegNode, string application, string processName = "")
         {
+            if (Process.GetCurrentProcess().ProcessName.Equals(processName, StringComparison.OrdinalIgnoreCase)) return;
```
droyad, line 26:
```suggestion
            if (Process.GetCurrentProcess().ProcessName.Equals(processName, StringComparison.OrdinalIgnoreCase)) 
                return;
```

source/Octopus.Server.Client/LocationChecker.cs
droyad, line 19: This is being removed as this part of the code only runs in the .NET Framework build of Octopus.Client and we don't ship that with server since we converted to .NET.

... Or do we, I'll follow up.

droyad: Oh it is too! From the Octopus Server install directory
image

droyad: So we need to put this line back in (I was mistaken when I said we don't ship it with Server)

desruc:
Re-added the server check and made `processName` a required parameter.

droyad (approved):

PR #594, opened 2021-06-15
Adding VersionControlReference to DeployNewReleaseActionResource

MarkSiedle:
Relates to https://github.com/OctopusDeploy/OctopusDeploy/pull/9452

# Background

For Deploy new release we need to do the following to support Git:
* We will have two additional fields, one to select a git ref and one to select a channel from the git ref.
* We will then be able to get the list of environments made available by the lifecycle on the selected channel.

# Results

`DeployNewReleaseActionResource` has a new field: `VersionControlReference` to support selecting a Channel from a GitRef for VCS DeployNewRelease Actions


gupta-kartik (approved):
:shipit:

PR #595, opened 2021-06-16
Structured Inputs for StepPackageEndpointResource

egorpavlikhin:
Small fix to allow `StepPackageEndpointResource.Inputs` to be passed as an object (with an unknown set of properties). This is serialized into `DeploymentActionInputs` in Server.

Parent PR: https://github.com/OctopusDeploy/OctopusDeploy/pull/9082

### DIFF
--- a/source/Octopus.Client.Tests/Integration/Repository/MachineRepositoryTest.cs
+++ b/source/Octopus.Client.Tests/Integration/Repository/MachineRepositoryTest.cs
@@ -1,16 +1,21 @@
-using System;
-using System.IO;
-using System.Reflection;
+using System.Collections.Concurrent;
+using System.Linq;
 using Nancy;
+using Nancy.Extensions;
+using Newtonsoft.Json;
 using NUnit.Framework;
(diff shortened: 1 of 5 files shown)


source/Octopus.Client.Tests/Integration/Repository/MachineRepositoryTest.cs
```diff
@@ -1,16 +1,21 @@
...
 using Octopus.Client.Model;
+using Octopus.Client.Model.Endpoints;
 using Octopus.Client.Repositories.Async;
+using Octopus.Client.Serialization;
 
 namespace Octopus.Client.Tests.Integration.Repository
 {
     public class MachineRepositoryTest : HttpIntegrationTestBase
     {
+        private static ConcurrentDictionary<string, MachineResource> machines = new ConcurrentDictionary<string, MachineResource>();
```
egorpavlikhin, line 17: This isn't nice, but hopefully whoever uses this in the future notices that it's `Concurrent` for a reason.

PR #600, opened 2021-07-13
Use AzureSignTool for code signing

matt-richardson:
So that the cert is loaded from azure key vault rather than disk

outstanding tasks:
* [x] update teamcity to pass in new params (can we read from env?)
* [x] check we are still timestamping

### DIFF
--- a/build/Build.cs
+++ b/build/Build.cs
@@ -259,14 +259,14 @@ class Build : NukeBuild
         foreach (var file in files)
             arguments += $"\"{file}\" ";
 
-        AzureSignTool(arguments, customLogger: (_, message) => Logger.Normal(message));
+        AzureSignTool(arguments, customLogger: LogStdErrAsWarning);
     }
 
     void SignWithSignTool(AbsolutePath[] files, string url)
     {
         Logger.Info("Signing files using signtool.");
 
-        SignToolLogger = (_, message) => Logger.Normal(message);
+        SignToolLogger = LogStdErrAsWarning;
 
         SignTool(_ => _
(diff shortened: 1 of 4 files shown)


corey-underdown (approved):

build/Build.cs
```diff
@@ -199,14 +207,49 @@ class Build : NukeBuild
 
     void SignBinaries(AbsolutePath path)
     {
-        Info($"Signing binaries in {path}");
-        var files = path.GlobDirectories("**").SelectMany(x => x.GlobFiles("Octopus.*.dll"));
+        Logger.Info($"Signing binaries in {path}");
+        var files = path.GlobDirectories("**").SelectMany(x => x.GlobFiles("Octopus.*.dll")).ToArray();
+
+        if (string.IsNullOrEmpty(AzureKeyVaultUrl)
+            && string.IsNullOrEmpty(AzureKeyVaultAppId)
```
slewis74, line 214: This feels like these should be `||` rather than `&&`. If any of those things are blank then `SignWithAzureSignTool` isn't going to work, is it?

slewis74 (approved):
One minor question on the boolean logic inline. Overall looks 👍

build/Build.cs
matt-richardson: This is a direct clone of what's in server which has been working happily.
If any of these are blank, azuresigntool will error with a nice error message (theoretically at least), so I'm not massively concerned here.

slewis74: Fair enough 👍

corey-underdown (approved):

PR #604, opened 2021-07-29
send the expected payload to config as code channel create and update…

LeeMeyer:
… controllers

### DIFF
--- a/source/Octopus.Server.Client/Repositories/Async/ChannelRepository.cs
+++ b/source/Octopus.Server.Client/Repositories/Async/ChannelRepository.cs
@@ -62,7 +62,7 @@ namespace Octopus.Client.Repositories.Async
     public interface IChannelBetaRepository
     {
         Task<ChannelResource> Get(ProjectResource projectResource, string idOrHref, string gitRef = null);
-        Task<ChannelResource> Create(ProjectResource projectResource, ChannelResource channelResource, string gitRef = null);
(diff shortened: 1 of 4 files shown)


source/Octopus.Server.Client/Repositories/Async/ChannelRepository.cs
```diff
@@ -90,15 +90,23 @@ public async Task<ChannelResource> Get(ProjectResource projectResource, string i
...
             var link = projectResource.Link("Channels");
-            return await client.Create(link, channelResource, new { gitRef });
+
+            var commitResouce = new CommitResource<ChannelResource>
+            {
+                CommitMessage = commitMessage,
+                Resource = channelResource
+            };
+
+            return await client.Post<CommitResource, ChannelResource>(link, commitResouce, new { gitRef });
```
emilol, line 109: I notice our other Beta repos use `Put` ([example](https://github.com/OctopusDeploy/OctopusClients/blob/master/source/Octopus.Server.Client/Repositories/DeploymentProcessRepository.cs#L71)) - is this endpoint wired up the same way? Might be worth running a test of the client in linqpad to confirm.

LeeMeyer: @emilol but that one is modify and this one is create (I might be missing something)?

emilol: ah, my bad! Nope not you, I missed that this was a create 😅👍

emilol (approved):

PR #605, opened 2021-07-30
Remove GetAllRunbooks operation from beta repository

dylanlerch:
Runbooks are not yet version controlled, so the route for this operation has been removed. We'll re-add this in the future once we migrate Runbooks to version control.


emilol (approved):
LGTM 👍

source/Octopus.Server.Client/Repositories/ProjectRepository.cs
```diff
@@ -124,10 +124,6 @@ public ResourceCollection<RunbookResource> GetRunbooks(ProjectResource project,
 
         public IReadOnlyList<RunbookResource> GetAllRunbooks(ProjectResource project)
         {
-            if (project.PersistenceSettings is VersionControlSettingsResource)
```
dylanlerch, line 127: This check was only in the sync repository.

PR #608, opened 2021-08-05
Add modify commands for version-controlled resources

emilol:
and consolidate commit/non-commit requests into the same resource/command. 

This improves the migration story for anyone converting to source control, and lowers our maintenance burden by getting us one step closer to having consolidated endpoints for project-scoped routes that may or may not include a `{gitRef}`.

# Background

[Trello card](https://trello.com/c/vbcFEmA5/359-commit-message-middleware) (internal link), relates to https://github.com/OctopusDeploy/OctopusDeploy/pull/9735

# Results

These changes are backwards-compatible with our existing DB routes, but are a breaking change for our `{gitRef}`'d routes, because they remove the need for a `CommitResource` wrapper around Create and Modify requests.

## Before

API consumers would need to a branching code-path to deal with commit messages based on if `projectResource.IsVersionControlled`:

```csharp
public async Task<DeploymentProcessResource> Modify(ProjectResource projectResource, DeploymentProcessResource resource, string commitMessage = null)
{
    if (!projectResource.IsVersionControlled)
    {
        // DB requests get sent as a resource
        return await client.Update(projectResource.Link("DeploymentProcess"), resource);
    }

    // we change the shape of requests for Git resources
    var commitResource = new CommitResource<DeploymentProcessResource>
    {
        Resource = resource,
        CommitMessage = commitMessage
    };
    await client.Update(resource.Link("Self"), commitResource);
    return await client.Get<DeploymentProcessResource>(resource.Link("Self"));
}
```

## After

The same Modify/Create command can be used regardless of whether the resource is stored in got or not:

```csharp
public async Task<ChannelResource> Create(ProjectResource projectResource, ModifyChannelCommand command, string gitRef = null)
{
    if (!(projectResource.PersistenceSettings is VersionControlSettingsResource settings))
        // we send a command for DB
        return await repository.Channels.Create(projectResource, command);

    gitRef = gitRef ?? settings.DefaultBranch;

    var link = projectResource.Link("Channels");

    // same command for Git
    return await client.Create(link, command, new { gitRef });
}
```

### DIFF
--- a/source/Octopus.Client.Tests/PublicSurfaceAreaFixture.ThePublicSurfaceAreaShouldNotRegress..NETCore.approved.txt
+++ b/source/Octopus.Client.Tests/PublicSurfaceAreaFixture.ThePublicSurfaceAreaShouldNotRegress..NETCore.approved.txt
@@ -2300,6 +2300,18 @@ Octopus.Client.Model
     static System.String StepName
     static Dictionary<String, String> AsDisplaySettings(String)
   }
+  class CreateChannelCommand
+    Octopus.Client.Extensibility.IResource
+    Octopus.Client.Model.IAuditedResource
+    Octopus.Client.Extensibility.INamedResource
+    Octopus.Client.Extensibility.IHaveSpaceResource
+    Octopus.Client.Model.IHaveProject
+    Octopus.Client.Model.ICommitCommand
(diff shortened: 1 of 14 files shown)


zentron:
LGTM

PR #609, opened 2021-08-06
Make naming consistent with async repo

emilol:
Does what is says on the tin. Sync was `IDeploymentProcessRepositoryBeta`, async was `IDeploymentProcessBetaRepository`, now they're both `IDeploymentProcessBetaRepository`


dylanlerch (approved):
LGTM 👍

PR #616, opened 2021-09-21
Remove channels from VCS

dylanlerch:


dylanlerch:
Build was broken on a package push conflict because I rebuilt the same commit

PR #617, opened 2021-10-15
Added AllowDefaultProxy to support specifying that the default proxy …

MissedTheMark:
…should or shouldn't be used.  Defaults to true, to maintain compatibility with previous versions


source/Octopus.Server.Client/OctopusAsyncClient.cs
```diff
@@ -45,6 +45,7 @@ protected OctopusAsyncClient(OctopusServerEndpoint serverEndpoint, OctopusClient
             {
                 CookieContainer = cookieContainer,
                 Credentials = serverEndpoint.Credentials ?? CredentialCache.DefaultNetworkCredentials,
+                UseProxy = clientOptions.AllowDefaultProxy
```
MissedTheMark, line 48: UseProxy defaults to true, so by specifying it explicitly here, we can disable use of the default proxy.  If a proxy is specified, this will be set to 'true' later in this method.

PR #619, opened 2021-11-12
Add shared connection store

dylanlerch:


dylanlerch:
Store is going to be Git-specific for now: https://github.com/OctopusDeploy/OctopusClients/pull/620

PR #631, opened 2022-02-08
Upgraded to Nuke 6.0.1

corey-underdown:


matt-richardson (approved):
Change looks good to me. Looks like some nuget failures in the build though - I've re-triggered.
I'd recommend running this by #team-integrations-fnm to keep them in the loop.

PR #633, opened 2022-02-12
Changing Permission to not be an enum

slewis74:
The Permission enum causes us issues every time a new permission is added to server. Older versions of client talking to newer versions of server with new permissions cause the Json deserialization to throw an exception when it doesn't recognise an enum value. This is the nature of enums.

In Octopus server itself, and in our other client libraries, we have moved most things away from using enums now. The preferred option over enums is a class that represents the same value but internally is just the string value off the wire.

To show some examples, the AdministerSystem permission is now defined as follows on the permissions class.

```
public static readonly Permission AdministerSystem = new Permission("AdministerSystem");
```

The Permission class itself is implementing `IEquatable` and has a `ToString` override, so for all intents and purposes it should still act exactly like an enum when you're consuming it.

E.g. if you had any of the following pieces of code before this change they would still compile after it
```
if (someObject.Permission == Permission.AdministerSystem)
```
```
if (someObject.Permissions.Contains(Permission.AdministerSystem))
```
```
someObject.Permissions.Add(Permission.AdministerSystem);
```

A new PermissionConverter is included and is wired in by default. It simply does a `ToString` of the object instances to serialize them and creates a new instance with the string value to deserialize. This means that if server sends a new value that didn't exist at client's compile time then there'll just be an object instance that won't equate to any known Permission instances, but it won't throw any exceptions.

This PR is likely easier to review per commit.

### DIFF
--- a/source/Octopus.Server.Client/Serialization/PermissionConverter.cs
+++ b/source/Octopus.Server.Client/Serialization/PermissionConverter.cs
@@ -1,0 +1,42 @@
+﻿#nullable enable
+using System;
+using Newtonsoft.Json;
+using Octopus.Client.Model;
+
+namespace Octopus.Client.Serialization
+{
+    public class PermissionConverter : JsonConverter
+    {
+        public override bool CanConvert(Type objectType)
+        {
+            return objectType == typeof(Permission);
+        }
+
+        public override void WriteJson(JsonWriter writer, object? value, JsonSerializer serializer)
+        {
+            if (value == null)
+            {
+                writer.WriteNull();
+                return;
+            }
+
+            var v = value.ToString();
+            writer.WriteValue(v);
+        }
+
+        public override object? ReadJson(JsonReader reader, Type objectType, object? existingValue, JsonSerializer serializer)
+        {
+            if (reader.TokenType == JsonToken.Null)
+            {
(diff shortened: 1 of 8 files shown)


source/Octopus.Server.Client/Serialization/PermissionConverter.cs
```diff
@@ -0,0 +1,42 @@
...
+        public override object? ReadJson(JsonReader reader, Type objectType, object? existingValue, JsonSerializer serializer)
+        {
+            if (reader.TokenType == JsonToken.Null)
+            {
+                return null;
+            }
+
+            if (reader.TokenType == JsonToken.String)
+            {
+                return new Permission(reader.Value!.ToString());
```
uglybugger, line 36: This looks like the meat of the change; correct?

uglybugger (approved):
This looks like a sensible change to me. Adding value to enums has bitten us so many times already...

source/Octopus.Server.Client/Serialization/PermissionConverter.cs
slewis74: Yeah, this is the meat of keeping what's on the wire the same.

PR #634, opened 2022-02-14
Keep all Octopus types public when ILMerging

### LINKED ISSUE #635: ILMerging Octopus.Client marks public interfaces as internal.
# Background

We publish two versions of the Octopus client:
1. An ILMerged version to enable API clients to reference a single `Octopus.Client.dll` file; and
2. A non-ILMerged version (requiring multiple NuGet dependencies) for application-level usage.

The former is intended for use in scenarios like PowerShell/Bash scripts where a single binary is helpful. The latter is intended for use in C# client applications where a normal NuGet package restore is more viable.

In general terms, these two packages are intended to be functionally identical - the `Octopus.Client.dll` assembly is literally just an ILMerged version of `Octopus.Server.Client.dll` and all of its `Octopus.*.dll` dependencies.

# Issue

Unfortunately, an existing configuration setting in the ILMerge operation is set to mark `public` interfaces as `internal` unless they are top-level interfaces directly declared in the `Octopus.Server.Client` assembly - meaning that some interfaces from `Octopus.Server.MessageContracts.Base` which are intended to be `public` are in fact being marked as `internal` - but only in the ILMerged version of the client. This is a parity mismatch which we need to correct.

We have automated tests which take dependencies on the ILMerged client and (successfully) instantiate and use it. What actually goes 💥 is where someone creates a mock of `IOctopusAsyncClient` using something like NSubstitute, e.g. `var client = Substitute.For<IOctopusAsyncClient>();`. This fails when NSubstitute attempts to dynamically create a method implementation for `IOctopusAsyncClient.Do<TCommand, TResponse>(ICommand command, CancellationToken cancellationToken)` because the method signature is `public` and yet the parameter type has been rewritten to be `internal`.

uglybugger:
# Background

Fixes #635.

# Results

This PR ensures that all relevant Octopus-defined types keep their existing visibility (i.e. `public`) when being ILMerged into a single `Octopus.Client.dll`.

It also works around an apparent bug in `il-repack` which ignores type name exclusions.

# Testing this PR

1. Run up a new C# solution. A simple console app is fine.
2. Restrict the package sources to nuget.org and `Install-Package Octopus.Client`
3. Observe that the `ICommand<,>` interface is marked as internal. (Try to define a type which implements it and it will go red pretty quickly.)
4. Remove the package reference.
5. Restrict the package sources to the Octopus internal NuGet feed and `Install-Package Octopus.Client -Pre -Version 11.4.3622-bug-ilmerging-internalises-public-interfaces`
6. Observe that the test code now compiles.

### DIFF
--- a/build/Build.cs
+++ b/build/Build.cs
@@ -44,15 +44,15 @@ class Build : NukeBuild
     AbsolutePath OctopusClientFolder => SourceDir / "Octopus.Client";
     AbsolutePath OctopusNormalClientFolder => SourceDir / "Octopus.Server.Client";
 
-    [Parameter("Whether to auto-detect the branch name - this is okay for a local build, but should not be used under CI.")] 
+    [Parameter("Whether to auto-detect the branch name - this is okay for a local build, but should not be used under CI.")]
     readonly bool AutoDetectBranch = IsLocalBuild;
-    
+
     [Parameter("Branch name for OctoVersion to use to calculate the version number. Can be set via the environment variable " + CiBranchNameEnvVariable + ".", Name = CiBranchNameEnvVariable)]
     string BranchName { get; set; }
 
-    [OctoVersion(BranchParameter = nameof(BranchName), AutoDetectBranchParameter = nameof(AutoDetectBranch))] 
+    [OctoVersion(BranchParameter = nameof(BranchName), AutoDetectBranchParameter = nameof(AutoDetectBranch))]
     public OctoVersionInfo OctoVersionInfo;
-    
+
     [PackageExecutable(
         packageId: "azuresigntool",
         packageExecutable: "azuresigntool.dll")]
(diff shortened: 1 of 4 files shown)


build/Build.cs
```diff
@@ -99,51 +99,76 @@ class Build : NukeBuild
...
+                    .SetLib(inputFolder));
 
                 DeleteDirectory(inputFolder);
                 MoveDirectory(outputFolder, inputFolder);
             }
         });
 
+    Target Test => _ => _
+        .DependsOn(Compile)
+        .DependsOn(Merge)   // IMPORTANT: Tests must be run _after_ the merge so that we're confident that we're testing the ILMerged code.  -andrewh 14/2/2022.
```
LukeButters, line 160: Is it possible to assert in test that we are indeed running on the merged code perhaps we could detect that some dll is indeed no loaded.

uglybugger: I think there's probably value in that. We already have assertions on whether there are certain .nuspec file references (i.e. there shouldn't be) and the thing still has to link before the tests will run, so we have an _implicit_ test for that, but it would be good to test it explicitly.

uglybugger: I'll make a note of that to follow up with in a separate PR given that this issue is currently blocking a customer.

LukeButters (approved):

MarkSiedle (approved):
Thank you. LGTM

PR #638, opened 2022-02-15
Make Octopus.TinyTypes.* internal again when ILMerging.

uglybugger:
# Background

We recently adjusted ILMerge settings to ensure that types from `Octopus.Server.MessageContracts.Base` and `Octopus.Server.MessageContracts.Base.HttpRoutes` remained public after an ILMerge. We brought all of the Octopus.* types along with that change, only to discover that there are some usages where there is now a conflict with some users' references to `Octopus.TinyTypes.*`.

Ultimately, the single ILMerged client will need some more invasive surgery before it can support API vNext payloads. Until then, though, its knowledge of Octopus.TinyTypes.* can (and should) be kept internal.

### DIFF
--- a/build/Build.cs
+++ b/build/Build.cs
@@ -118,8 +118,12 @@ class Build : NukeBuild
                 // 2) that temporary assembly plus all of the third-party assemblies, leaving only the types from the first (Octopus temporary) assembly as public.
                 // --andrewh 14/2/2022.
 
-                // Stage 1: Merge all the Octopus assemblies.
-                var stage1Assemblies = inputFolder.GlobFiles("Octopus.*.dll")
+                // Stage 1: Merge all the Octopus assemblies whose contracts we want to not internalize.
+                var stage1Assemblies = inputFolder.GlobFiles(
+                        "Octopus.Server.Client.dll",
+                        "Octopus.Server.MessageContracts.Base.dll",
+                        "Octopus.Server.MessageContracts.Base.HttpRoutes.dll"
+                    )
                     .Select(x => x.ToString())
                     .OrderBy(x => x)
                     .ToArray();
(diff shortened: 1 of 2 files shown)


build/Build.cs
```diff
@@ -118,8 +118,12 @@ class Build : NukeBuild
...
                 // --andrewh 14/2/2022.
 
-                // Stage 1: Merge all the Octopus assemblies.
-                var stage1Assemblies = inputFolder.GlobFiles("Octopus.*.dll")
+                // Stage 1: Merge all the Octopus assemblies whose contracts we want to not internalize.
+                var stage1Assemblies = inputFolder.GlobFiles(
+                        "Octopus.Server.Client.dll",
+                        "Octopus.Server.MessageContracts.Base.dll",
+                        "Octopus.Server.MessageContracts.Base.HttpRoutes.dll"
+                    )
```
uglybugger, line 126: This is the meat of this change: be specific about the assemblies we want to keep public.

source/Octopus.Client.E2ETests/InlinedDependenciesFixture.cs
```diff
@@ -48,23 +44,42 @@ public void OneTimeTearDown()
...
+            {
+                case Visibility.Internal:
+                    type!.IsPublic.Should().BeFalse();
+                    break;
+                case Visibility.Public:
+                    type!.IsPublic.Should().BeTrue();
+                    break;
+                default:
+                    throw new ArgumentOutOfRangeException(nameof(expectedVisibility), expectedVisibility, null);
+            }
```
uglybugger, line 76: Added assertions about which types are ILMerged as `public` versus `internal`.

matt-richardson (approved):
Nice one. After reading through this and the [other pr](https://github.com/OctopusDeploy/OctopusClients/pull/634/files), I think this will get us to the right place 👍 

I'd still recommend someone from Core look over it as well, just in case I'm missing some context.

PR #639, opened 2022-02-17
Add IsPrivate field to SpaceResource

iryan2:
Related to work in [this pr](https://github.com/OctopusDeploy/OctopusDeploy/pull/11299)

### DIFF
--- a/source/Octopus.Server.Client/Model/SpaceResource.cs
+++ b/source/Octopus.Server.Client/Model/SpaceResource.cs
@@ -20,6 +20,8 @@ namespace Octopus.Client.Model
         [Writeable]
         public bool IsDefault { get; set; }
 
+        public bool IsPrivate { get; set; }
+
         public bool TaskQueueStopped { get; set; }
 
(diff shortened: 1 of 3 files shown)


source/Octopus.Server.Client/Model/SpaceResource.cs
```diff
@@ -20,6 +20,8 @@ public SpaceResource()
         [Writeable]
         public bool IsDefault { get; set; }
 
+        public bool IsPrivate { get; set; }
```
iryan2, line 23: I left the `Writeable` attribute off `IsPrivate` in the guess that that's the right move, according to the logic in [this comment](https://github.com/OctopusDeploy/OctopusDeploy/pull/11381/files#r806482691). Is this correct?

andrewabest: Yes that would be the right thing to do.

andrewabest (approved):
LGTM

PR #646, opened 2022-03-11
Realigning Build Information resources for LTS 2021.3

slewis74:
The client resources for Deployment Changes haven't stayed in alignment with those that Octopus server is actually using, this brings them back into alignment.

Relates to OctopusDeploy/Issues#7425


slewis74:
Replaced by #647

PR #649, opened 2022-03-18
Add PrivateSpaceOwner to SpaceResource

iryan2:
Related to work in this [PR](https://github.com/OctopusDeploy/OctopusDeploy/pull/11641)

### DIFF
--- a/source/Octopus.Client.Tests/PublicSurfaceAreaFixture.ThePublicSurfaceAreaShouldNotRegress..NETCore.approved.txt
+++ b/source/Octopus.Client.Tests/PublicSurfaceAreaFixture.ThePublicSurfaceAreaShouldNotRegress..NETCore.approved.txt
@@ -4900,6 +4900,7 @@ Octopus.Client.Model
     Boolean IsDefault { get; set; }
     Boolean IsPrivate { get; set; }
     String Name { get; set; }
+    String PrivateSpaceOwner { get; set; }
     Octopus.Client.Model.ReferenceCollection SpaceManagersTeamMembers { get; set; }
     Octopus.Client.Model.ReferenceCollection SpaceManagersTeams { get; set; }
     Boolean TaskQueueStopped { get; set; }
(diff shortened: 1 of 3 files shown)


iryan2:
@andrewabest Is there anything else we should be including in this change? I hadn't planned on doing this immediately, but I believe I need it for the change to the [test we spoke about](https://github.com/OctopusDeploy/OctopusDeploy/pull/11448/files#r827614256), or at least doing it up front will prevent a temporary workaround.

andrewabest:
> @andrewabest Is there anything else we should be including in this change? I hadn't planned on doing this immediately, but I believe I need it for the change to the [test we spoke about](https://github.com/OctopusDeploy/OctopusDeploy/pull/11448/files#r827614256), or at least doing it up front will prevent a temporary workaround.

No this seems fine. It isn't a writeable property, which we want, and it will either be null if no value exists, or have the appropriate value within it. Looks good.

source/Octopus.Client.Tests/PublicSurfaceAreaFixture.ThePublicSurfaceAreaShouldNotRegress..NETCore.approved.txt
```diff
@@ -4900,6 +4900,7 @@ Octopus.Client.Model
     Boolean IsDefault { get; set; }
     Boolean IsPrivate { get; set; }
```
andrewabest, line 4901: I suspect there is also the netframework approval, not sure how you generated that last time though.

andrewabest (approved):
Looks good

source/Octopus.Client.Tests/PublicSurfaceAreaFixture.ThePublicSurfaceAreaShouldNotRegress..NETCore.approved.txt
iryan2: Yeah, [last time](https://github.com/OctopusDeploy/OctopusClients/pull/639) I updated two approved files, this time I was only prompted for one. The test passes whether I include it or not, and it's passed all the tests here. Should I be concerned or just go ahead and merge it?

andrewabest: Last time you wouldn't have been prompted for the netframework file - you are on a mac, and so can't run that test. You must have updated it by hand. I would do the same here, update the netframework file by hand to ensure it matches what we would expect to be the output.

iryan2: Ah sure, I guess it was a CI failure that made me update it last time. Which begs the question, why did it pass without it? Anyway, I've updated it it's passed again so I suppose I'll merge this.

PR #650, opened 2022-03-23
Fall back to ConvertToVcs link if there is no ConvertToGit link

dylanlerch:
We changed the name of the link that's used when converting projects to Git (from `ConvertToVcs` to `ConvertToGit`). This is part of a longer term goal to change all of our usages of `Vcs` to `Git`.

To make this transition as smooth as possible, the projects links collection for a project contains both of these links for the time being, so older versions of client would continue to work with newer versions of Octopus.

But I didn't cater for using a newer version of client with an older version of Octopus. Client is looking for `ConvertToGit` but older versions of Octopus are only returning the `ConvertToVcs` link. This closes that loop, falling back to the old link if the new one is not there.

### DIFF
--- a/source/Octopus.Client.Tests/Repositories/ProjectRepositoryFixture.cs
+++ b/source/Octopus.Client.Tests/Repositories/ProjectRepositoryFixture.cs
@@ -1,0 +1,93 @@
+using System.Threading.Tasks;
+using FluentAssertions;
+using NSubstitute;
+using NUnit.Framework;
+using Octopus.Client.Extensibility;
+using Octopus.Client.Model;
+using Octopus.Client.Model.Git;
+
+namespace Octopus.Client.Tests.Repositories
+{
+    [TestFixture]
+    public class ProjectRepositoryFixture
+    {
+        OctopusAsyncRepository asyncRepository;
+        string asyncUrlUsed;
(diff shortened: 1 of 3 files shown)


source/Octopus.Client.Tests/Repositories/ProjectRepositoryFixture.cs
```diff
@@ -0,0 +1,90 @@
...
+                    {"ConvertToVcs", new Href("/convert/to/vcs")}
+                }
+            };
+
+            // Act
+            await asyncRepository.Projects.ConvertToGit(project, new GitPersistenceSettingsResource(), string.Empty);
+            syncRepository.Projects.ConvertToGit(project, new GitPersistenceSettingsResource(), string.Empty);
+
+            // Assert
+            asyncUrlUsed.Should().Be("/convert/to/git");
```
LeeMeyer, line 50: hardcoded this path in two places? is it worth having a constant?

LeeMeyer (approved):
all seems sensible to me. only suggestion that came to my mind was to move the hardcoded paths to constants to make things a bit easier to maintain, maybe

StephenHeise (approved):
LGTM.

PR #656, opened 2022-05-26
Add GitRef parameter to VariablePreview

dylanlerch:


YuKitsune (approved):
Should be good 👍 
I've noticed that our other methods (deployment process/settings) have a separate method with the `gitRef` argument instead of having an optional argument. 
Might be good to have a separate overload just for the `gitRef` argument, but I'll leave that up to you

PR #657, opened 2022-05-29
Add VariablesGitReference to Release and RunbookSnapshot GitReferences

dylanlerch:


source/Octopus.Server.Client/Model/RunbookSnapshotGitReferenceResource.cs
```diff
@@ -0,0 +1,8 @@
+namespace Octopus.Client.Model
+{
+    public class RunbookSnapshotGitReferenceResource
```
YuKitsune, line 3: Should this inherit from `GitReferenceResource`, or is it intentionally separate?

dylanlerch: Intentionally separate. Ideally `ReleaseGitReferenceResource` would also be separate, but it's inheriting at the moment for backwards compatibility reason. When we major bump when we release variables, we'll remove that inheritance.

Reason for not inheriting here is that that currently doesn't have a `GitCommit` value, because variables are the only thing that come from a commit.

YuKitsune (approved):

source/Octopus.Server.Client/Model/RunbookSnapshotGitReferenceResource.cs
dylanlerch: So we ended up doing exactly as you said. The reason for keeping them separate was because we didn't need the `GitCommit` field on the Runbook snapshots just yet. But we'll need it shortly when we migrate Runbooks, so seemed pointless to have separate models for a few months. `GitCommit` on the Runbook snapshot will just be `null`.

PR #665, opened 2022-07-13
Added depends on tests to build, made framework approved public surface file match core version

stevencl840:
While doing preparation work for the .Net 6 migration it was discover we have a failing unit test (ThePublicSurfaceAreaShouldNotRegress) when testing .Net Framework. The failing test was investigated to discover how the build pipeline was functioning if this was not succeeding.

Discoveries
* The unit tests can only run against .Net Framework in a Windows environment
* The units tests were not being called in the Windows environment
* The unit tests were being called in various Linux environments and passing
* Changes that had been made to the Approved Public Surface for .Net Core had not been made to .Net Framework

This PR seeks to solve the problem by
* Including the unit tests in the Build process, to ensure they are run in the build environment(Windows)
* Updated the Approved Public Surface for .Net Framework to include the required updates, these changes are already present in the .Net Core version of the Approved Public Surface.

### DIFF
--- a/build/Build.cs
+++ b/build/Build.cs
@@ -257,10 +257,10 @@ class Build : NukeBuild
         var files = path.GlobDirectories("**").SelectMany(x => x.GlobFiles("Octopus.*.dll")).ToArray();
 
         var useSignTool = string.IsNullOrEmpty(AzureKeyVaultUrl)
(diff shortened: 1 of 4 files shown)


skermajo (approved):
TYVM 👍

PR #670, opened 2022-07-25
Remove IsJiraServiceManagementIntegrationEnabled flag

rain-on:
The JSM Deployment Blocking Integration is now protected via a licence flag, and no longer requires a feature flag - as such, the flag is being removed.


benPearce1 (changes requested):
You will need to remove the entries in the approval files as per https://github.com/OctopusDeploy/OctopusClients/pull/667

benPearce1 (approved):
LGTM

PR #673, opened 2022-07-27
Adding event retention resource and API interfaces

sburmanoctopus:
# Background
The new Event Retention feature was added to Octopus Server. We now want to update the Octopus.Client to reflect the enhancements made.

# Results
We now support the 3 endpoints for managing archived event files in the Octopus.Client.

The configuration resource was also added to support using the configuration endpoint with this new feature.

## Testing
Octopus Server was updated using the NuGet Package of this PR, and enhanced to use the new methods.

The PR can be found [here](https://github.com/OctopusDeploy/OctopusDeploy/pull/13135)

Is this considered an acceptable level of testing for this change?

# How to Review this PR
This is the first time I have made a change to Octopus.Client. Please make sure I have not overlooked something that needs to be set.

# Before we Merge
The minimum version required has not been built yet. So this PR cannot be merged until this has been determined and this PR has updated the `ArchivedEventFileRepository` (both sync and async) with this version.

### DIFF
--- a/source/Octopus.Server.Client/Model/EventRetention/EventRetentionConfigurationResource.cs
+++ b/source/Octopus.Server.Client/Model/EventRetention/EventRetentionConfigurationResource.cs
@@ -1,0 +1,21 @@
+﻿using Octopus.Client.Extensibility;
+using Octopus.Client.Extensibility.Attributes;
+
+namespace Octopus.Client.Model.EventRetention
+{
+    public class EventRetentionConfigurationResource : IResource
+    {
+
+        public EventRetentionConfigurationResource()
+        {
+            Id = "eventretention";
+        }
+
+        public string Id { get; set; }
+        
+        [Writeable]
+        public int EventRetentionDays { get; set; } = 90;
+
+        public LinkCollection Links { get; set; }
+    }
+}
--- a/source/Octopus.Server.Client/Repositories/Async/ArchivedEventFileRepository.cs
+++ b/source/Octopus.Server.Client/Repositories/Async/ArchivedEventFileRepository.cs
@@ -1,0 +1,45 @@
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Threading.Tasks;
+using Octopus.Client.Model;
+using Octopus.Client.Model.Endpoints;
+using Octopus.Client.Model.EventRetention;
+
+namespace Octopus.Client.Repositories.Async
+{
+    public interface IArchivedEventFileRepository : 
+        IDelete<ArchivedEventFileResource>
+    {
+        Task<Stream> GetContent(ArchivedEventFileResource archiveEventFile);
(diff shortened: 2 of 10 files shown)


source/Octopus.Server.Client/Model/EventRetention/EventRetentionConfigurationResource.cs
```diff
@@ -0,0 +1,21 @@
+﻿using Octopus.Client.Extensibility;
+using Octopus.Client.Extensibility.Attributes;
+
+namespace Octopus.Client.Model.EventRetention
+{
+    public class EventRetentionConfigurationResource : IResource
```
felixi-octopus, line 6: Any reason to inherit `IResource` over `Resource`?

felixi-octopus: OK. Is `MaintenanceConfigurationResource` an inconsistency? I was using that as a reference

source/Octopus.Server.Client/Repositories/Async/ArchivedEventFileRepository.cs
```diff
@@ -0,0 +1,45 @@
...
+    {
+        Task<Stream> GetContent(ArchivedEventFileResource archiveEventFile);
+        Task<ResourceCollection<ArchivedEventFileResource>> List(int skip = 0, int? take = null);
+    }
+
+    class ArchivedEventFileRepository : BasicRepository<ArchivedEventFileResource>, IArchivedEventFileRepository
+    {
+        public ArchivedEventFileRepository(IOctopusAsyncRepository repository) : base(repository, "ArchivedEventFiles")
+        {
+            MinimumCompatibleVersion("2022.3.6072");
```
felixi-octopus, line 22: Are release version are we awaiting here? GA as this is a public repo?

felixi-octopus (approved):
LGTM. Awaiting version change

source/Octopus.Server.Client/Model/EventRetention/EventRetentionConfigurationResource.cs
sburmanoctopus: Huh, so it does 🙂 I don't know why sorry. 

I can't find any usage of the `LastModified...` fields for `MaintenanceConfigurationResource` (other than a mapper, but that was auto-generated). So I suspect it's not important.

source/Octopus.Server.Client/Repositories/Async/ArchivedEventFileRepository.cs
sburmanoctopus: This will be the version we use when rolling out this feature. 

In the plan we merge this PR after we turn this feature on for some pilot customers. So whatever version we are targeting at that point.

PR #687, opened 2022-08-16
s/InsightsWorkspace/InsightsReport

MissedTheMark:


droyad (approved):
No-one should have been using these as they don't work on Octopus Server, so 👍

PR #691, opened 2022-08-22
Fix - Nightly build produces branch TAG

michaelongso:
For the nightly build, the branch information is `refs/heads/main` and the existing expression is appending `-nightly`.
However, `refs/heads/main-nightly`, is matching the `NonPreReleaseTagsRegex` defined in `octoversion.json`
This results in **no pre-release** tag added for nightly build. 

The original regex captures any additional character / words beyond the condition (main|master...*), which is not correct.


stevencl840 (approved):
LGTM

PR #681, opened 2022-08-15
IOctopusAsyncClient cancellation token support

michaelongso:
Adding cancellation token support for `IOctopusAsyncClient`, `IVariableSetRepository`  and `IReleaseRepository`.

All existing methods that doesn't take cancellation token are marked with `Obsolete`

The guidance on adding cancellation token overload is to maintain compliance with [**CA1068: CancellationToken parameters must come last**](https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1068)

PR #702, opened 2022-09-15
Add cancellation token support for IDeploymentProcessRepository

michaelongso:
This is a continued effort on initial initiative completed in this [PR](https://github.com/OctopusDeploy/OctopusClients/pull/681).

For this stream, the cancellation token support is added on `IDeploymentProcessRepository`.

All existing methods that doesn't take cancellation token are marked with `Obsolete`

The guidance on adding cancellation token overload is to maintain compliance with **[CA1068: CancellationToken parameters must come last](https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1068)**


MarkSiedle (approved):
Great to see this spreading 👍 LGTM

PR #706, opened 2022-10-07
Updated versio of MessageContracts.Base and MessageContracts.Base.HttpRoutes

stevencl840:
The previous version of the Client was targeting a version of Octopus.Server.MessageContracts.Base that was not publicly  accessible. Octopus.Client has the version of the message contracts merged so was not affected. But Octopus.Server.Client does not have it's dependencies merged so could not be used.


akirayamamoto (approved):
LGTM. Thanks Steven.
Context: https://octopusdeploy.slack.com/archives/C01HZFJRYSH/p1665070515338779

PR #711, opened 2022-11-15
Add ApiKeysEnabled

N-lson:
[sc-31164]


source/Octopus.Server.Client/Model/AuthenticationConfigResource.cs
```diff
@@ -35,5 +35,8 @@ public AuthenticationConfigResource()
 
         [Writeable]
         public bool RememberMeEnabled { get; set; }
+
+        [Writeable]
+        public bool ApiKeysEnabled { get; set; }
```
tothegills, line 40: The year is 2023. Secops want to disable API keys, not just for service accounts. What is that new property called?

N-lson: 👍 Changed to `ApiKeysEnabledForNonServiceAccounts`

source/Octopus.Server.Client/Model/AuthenticationConfigResource.cs
```diff
@@ -35,5 +35,8 @@ public AuthenticationConfigResource()
 
         [Writeable]
         public bool RememberMeEnabled { get; set; }
+
+        [Writeable]
+        public bool ApiKeysEnabledForNonServiceAccounts { get; set; }
```
tothegills, line 40: Imagine we had 3 types of accounts: Apple, Blueberry and Watermelon.
The first requirement is to disable api keys for all but Watermelon.
```
ApiKeysEnabledForNonWatermelonAccounts
```

How would the code need to change in the following scenarios:
- The concept of Watermelon accounts is removed.
- A new requirement is added for finer-grained control: Apple or Blueberry accounts can be disabled.

N-lson: Are you suggesting to use something like `ForUserAccounts` instead of `ForNonXAccounts`? It would make it easier to change. Originally I went with NonService since I thought "User" could be a bit confusing. Service accounts are technically users as well. Not sure if there's a more descriptive name for "user accounts".

tothegills (approved):

PR #713, opened 2022-11-23
Set AssentNonInteractive for all build jobs

skermajo:
`AssentNonInteractive` was only being set on select build jobs. We need to set the value on all of them, so all the test runs pick them up.


stevencl840:
Will probably need to do this for a few other project too

stevencl840:
This look valid to me. I also cannot see how making this change would have a negative impact on the pipeline(provided we get a green build)

stevencl840 (approved):
This look valid to me. I also cannot see how making this change would have a negative impact on the pipeline(provided we get a green build)

PR #714, opened 2022-11-23
Obsoleted `UserPermissionsRepository.GetConfiguration` and replaced it with `GetDescriptions` which returns the right type

droyad:
[The API controller is here](https://github.com/OctopusDeploy/OctopusDeploy/blob/431e99f2f127352dfc137eef376444ce7a72df98/source/Octopus.Server/Web/Api/Actions/PermissionDefinitionsResponder.cs#L22)

While porting the API to ASP.NET (https://github.com/OctopusDeploy/OctopusDeploy/pull/15443) I found that the method didn't actually return any data because the response was in a completely different shape.

I decided to obsolete it to keep compat, but I doubt anyone is actually using it.

[sc-32326]

### DIFF
--- a/build/Build.cs
+++ b/build/Build.cs
@@ -102,7 +102,7 @@ class Build : NukeBuild
 
     Target Merge => _ => _
         .DependsOn(Compile)
-        .Executes(async () =>
+        .Executes(() =>
         {
             foreach (var target in new[] {"net462", "netstandard2.0"})
             {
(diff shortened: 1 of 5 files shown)


build/Build.cs
```diff
@@ -102,7 +102,7 @@ class Build : NukeBuild
 
     Target Merge => _ => _
         .DependsOn(Compile)
-        .Executes(async () =>
```
droyad, line 105: Fixes a warning during build

matt-richardson (approved):
Seems pretty logical to me. Crazy that it's never(?) worked.

Any sane way of adding a test in here for this?

droyad:
> Any sane way of adding a test in here for this?

Not that I can think of, we don't have integration tests with the server here.

PR #715, opened 2022-11-25
Don't overwrite genuine cancellation exceptions with timeout exceptions in DispatchRequest

rosslovas:
Regardless of what the actual cause of cancellation was, `DispatchRequest` squashes any cancellation exception in favour of a `TimeoutException`. This means if your own code makes a request and then cancels it immediately for whatever reason, you get the following misleading error:

```
System.TimeoutException: Timeout getting response from [...] (client timeout is set to 00:04:00).
```

This also means that user code that wants to squash a cancellation exception caused by their own token (e.g. during application shutdown, when cancellation should happen silently and not throw) cannot do so.

This PR:  
* Fixes that issue by specifically replacing the exception when it was genuinely caused by a timeout
* Keeps the original exception around as an inner exception as a matter of best practice
* Confirms the fix with a test

### DIFF
--- a/source/Octopus.Client.Tests/Integration/OctopusClient/TimeoutTests.cs
+++ b/source/Octopus.Client.Tests/Integration/OctopusClient/TimeoutTests.cs
@@ -4,7 +4,6 @@ using System.Threading;
 using System.Threading.Tasks;
 using FluentAssertions;
 using NUnit.Framework;
-using Octopus.Client.Exceptions;
 
 namespace Octopus.Client.Tests.Integration.OctopusClient
 {
@@ -36,5 +35,20 @@ namespace Octopus.Client.Tests.Integration.OctopusClient
             sw.Elapsed.Should().BeLessThan(TimeSpan.FromSeconds(10));
         }
 
+        [Test]
+        public void CancellationThrowsOperationCanceledException()
+        {
+            var sw = Stopwatch.StartNew();
+            var cancellationTokenSource = new CancellationTokenSource();
+
+            var getTask = AsyncClient.Get<string>("~/", cancellationTokenSource.Token);
+            cancellationTokenSource.Cancel();
+
+            var get = () => getTask;
+            get.ShouldThrow<OperationCanceledException>()
+                .Where(ex => ex.CancellationToken == cancellationTokenSource.Token);
+
+            sw.Elapsed.Should().BeLessThan(TimeSpan.FromSeconds(10));
+        }
     }
 }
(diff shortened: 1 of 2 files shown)


source/Octopus.Client.Tests/Integration/OctopusClient/TimeoutTests.cs
```diff
@@ -36,5 +35,20 @@ public void ConfiguredTimeoutWorks()
...
+            var cancellationTokenSource = new CancellationTokenSource();
+
+            var getTask = AsyncClient.Get<string>("~/", cancellationTokenSource.Token);
+            cancellationTokenSource.Cancel();
+
+            var get = () => getTask;
+            get.ShouldThrow<OperationCanceledException>()
+                .Where(ex => ex.CancellationToken == cancellationTokenSource.Token);
+
+            sw.Elapsed.Should().BeLessThan(TimeSpan.FromSeconds(10));
```
rosslovas, line 51: The stopwatch is not strictly necessary but it's something the above test `ConfiguredTimeoutWorks` also does, and it seems sensible to just verify that cancellation does the same.

source/Octopus.Server.Client/OctopusAsyncClient.cs
```diff
@@ -633,9 +633,11 @@ protected virtual async Task<OctopusResponse<TResponseResource>> DispatchRequest
                         return octopusResponse;
                     }
                 }
-                catch (TaskCanceledException)
+                catch (TaskCanceledException exception) when (
+                    exception.CancellationToken != cancellationToken &&
+                    exception.InnerException is TimeoutException)
                 {
-                    throw new TimeoutException($"Timeout getting response from {request.Uri} (client timeout is set to {client.Timeout}).");
+                    throw new TimeoutException($"Timeout getting response from {request.Uri} (client timeout is set to {client.Timeout}).", exception);
```
skermajo, line 640: This is a nice improvement. It's a shame the current design locks us into a TimeoutException wrapped in a TaskCanceledException wrapped in a TimeoutException.

source/Octopus.Server.Client/OctopusAsyncClient.cs
```diff
@@ -633,9 +633,11 @@ protected virtual async Task<OctopusResponse<TResponseResource>> DispatchRequest
                         return octopusResponse;
                     }
                 }
-                catch (TaskCanceledException)
+                catch (TaskCanceledException exception) when (
+                    exception.CancellationToken != cancellationToken &&
```
skermajo, line 637: Should/can this be `is not` instead of `!=`?

skermajo (approved):
Thanks, good focussed change 👍 Just one question for consideration.

source/Octopus.Server.Client/OctopusAsyncClient.cs
rosslovas: It can't unfortunately, pattern matching only supports compile time constants as values, which `cancellationToken` is not. IIRC it's being considered for a future C# version though 🙂

PR #719, opened 2022-12-06
Fix null exception in OctopusExceptionFactory

michaelongso:
[SC-33816] - [Details](https://octopusdeploy.slack.com/archives/C01HZFJRYSH/p1670290247361119)

# Summary
It was identified that `Null exception` will occur when response body is missing the expected `OctopusErrorsContract` structure.

# Result
- Refactored the code to have consistent handling when an exception occurs during JSON deserialization.
- Null check on deserialized `OctopusErrorsContract` object prior to accessing
- Unit tests as safety net


source/Octopus.Server.Client/Exceptions/OctopusExceptionFactory.cs
```diff
@@ -51,74 +51,95 @@ public static async Task<OctopusException> CreateException(HttpResponseMessage r
...
             }
 
-            if (statusCode == 401 || statusCode == 403) // Forbidden, usually no API key or permissions
+            if (statusCode is 401 or 403) // Forbidden, usually no API key or permissions
             {
-                var errors = JsonConvert.DeserializeObject<OctopusErrorsContract>(body);
-                errors = errors ?? new OctopusErrorsContract
-                         {
-                             ErrorMessage =
-                                 $"The server returned a status code of {statusCode}: {body}"
```
michaelongso, line 69: Refactored this error message. `OctopusSecurityException` contains the `statusCode`, having it repeated on the message is redundant. See [CreateOctopusSecurityException](https://github.com/OctopusDeploy/OctopusClients/pull/719/files#diff-e162142f5091b207e006db7032f222b1291bdbf6eb34d3bb761f4869dbc90fa6R77) and [OctopusErrorsContractFromBody](https://github.com/OctopusDeploy/OctopusClients/pull/719/files#diff-e162142f5091b207e006db7032f222b1291bdbf6eb34d3bb761f4869dbc90fa6R117)

source/Octopus.Server.Client/Exceptions/OctopusExceptionFactory.cs
```diff
@@ -51,74 +51,96 @@ public static async Task<OctopusException> CreateException(HttpResponseMessage r
 
         public static OctopusException CreateException(int statusCode, string body)
         {
-            if (statusCode == 400 || statusCode == 409) // Bad request: usually validation error 
+            if (statusCode is 400 or 409) // Bad request: usually validation error 
             {
-                var errors = JsonConvert.DeserializeObject<OctopusErrorsContract>(body);
-                return new OctopusValidationException(statusCode, errors.ErrorMessage, errors.Errors, errors.Details)
```
michaelongso, line 57: - `errors` can be null here
- OctopusValidationException will throw null exception when `errors.ErrorMessage` or `errors.Errors` is null

source/Octopus.Server.Client/Exceptions/OctopusExceptionFactory.cs
```diff
@@ -51,74 +51,96 @@ public static async Task<OctopusException> CreateException(HttpResponseMessage r
...
             }
-            catch
+            catch (Exception ex)
             {
-                errorMessage = body;
+                errorMessage = $"Unexpected json deserialization error for {body} :: {ex.Message}";
             }
-            return errorMessage;
+
+            return new OctopusErrorsContract
```
michaelongso, line 136: The sensible defaults are in place to prevent null exception when constructing `octopus exception` (IE: OctopusValidationException)

source/Octopus.Server.Client/Exceptions/OctopusExceptionFactory.cs
```diff
@@ -51,74 +51,96 @@ public static async Task<OctopusException> CreateException(HttpResponseMessage r
...
 
-            return new OctopusServerException(statusCode, fullDetails) {HelpText = helpText};
+            return new OctopusServerException(statusCode, fullDetails)
+            {
+                HelpText = errorsContract.HelpText
+            };
         }
 
-        static string GetErrorMessage(string body)
+        internal static OctopusErrorsContract OctopusErrorsContractFromBody(string body)
```
michaelongso, line 117: `internal` to support `expected result` for unit tests

source/Octopus.Client.Tests/Exceptions/OctopusExceptionFactoryFixture.cs
```diff
@@ -0,0 +1,304 @@
...
+            OctopusExceptionFactory.OctopusErrorsContract octopusErrorsContract)
+        {
+            var httpResponseMessage = new HttpResponseMessage(HttpStatusCode.BadRequest);
+
+            OctopusExceptionFactory.OctopusErrorsContract resultValidationObject = null;
+            if (HasAnyData(octopusErrorsContract))
+            {
+                var content = JsonConvert.SerializeObject(octopusErrorsContract);
+                httpResponseMessage.Content = new StringContent(content);
+                resultValidationObject = OctopusExceptionFactory.OctopusErrorsContractFromBody(content);
```
adam-mccoy, line 43: This uses the logic under test to generate the expected result. This should come from an independent source.

source/Octopus.Client.Tests/Exceptions/OctopusExceptionFactoryFixture.cs
```diff
@@ -0,0 +1,304 @@
...
+using System.Threading.Tasks;
+using FluentAssertions;
+using FluentAssertions.Execution;
+using Newtonsoft.Json;
+using NUnit.Framework;
+using Octopus.Client.Exceptions;
+
+namespace Octopus.Client.Tests.Exceptions
+{
+    public class OctopusExceptionFactoryFixture
```
adam-mccoy, line 13: I like that this fixture is very thorough!

source/Octopus.Server.Client/Exceptions/OctopusExceptionFactory.cs
```diff
@@ -51,74 +51,96 @@ public static async Task<OctopusException> CreateException(HttpResponseMessage r
...
+            }
+            catch (Exception jsonEx) when (jsonEx is ArgumentNullException or JsonSerializationException)
+            {
+                errorMessage = $"JSON deserialize error for {body} :: {jsonEx.Message}";
             }
-            catch
+            catch (Exception ex)
             {
-                errorMessage = body;
+                errorMessage = $"Unexpected json deserialization error for {body} :: {ex.Message}";
```
adam-mccoy, line 133: These error messages are worded slightly differently, but effectively say the same thing. Why the different messages depending on exception type?

If it were me, it'd probably have left the behaviour of this alone, and just stuck the body in the error message. I'm not sure it's helpful to potentially give customers exception messages about failed deserialization because the Octopus payload isn't valid. This information is probably better in a log message.

adam-mccoy (commented):
This is looking a lot better than it was. I've left a few suggestions for you. 🙇

source/Octopus.Server.Client/Exceptions/OctopusExceptionFactory.cs
```diff
@@ -51,74 +51,96 @@ public static async Task<OctopusException> CreateException(HttpResponseMessage r
 
         public static OctopusException CreateException(int statusCode, string body)
         {
-            if (statusCode == 400 || statusCode == 409) // Bad request: usually validation error 
+            if (statusCode is 400 or 409) // Bad request: usually validation error
```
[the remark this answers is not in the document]
michaelongso: Yes, much neater.

source/Octopus.Server.Client/Exceptions/OctopusExceptionFactory.cs
```diff
@@ -51,74 +51,96 @@ public static async Task<OctopusException> CreateException(HttpResponseMessage r
...
             string errorMessage;
+
             try
             {
-                var errors = JsonConvert.DeserializeObject<OctopusErrorsContract>(body);
-                errorMessage = errors.ErrorMessage;
+                result = JsonConvert.DeserializeObject<OctopusErrorsContract>(body);
+                errorMessage = result?.ErrorMessage ?? body;
+            }
+            catch (Exception jsonEx) when (jsonEx is ArgumentNullException or JsonSerializationException)
```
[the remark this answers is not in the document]
michaelongso: I'm contemplating on just relying on the `catch (Exception ex)`. Will think it through. Thanks!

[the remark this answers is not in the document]
michaelongso: Decided to simplified it as a single catch exception block

source/Octopus.Client.Tests/Exceptions/OctopusExceptionFactoryFixture.cs
adam-mccoy: I don't believe this is actually testing anything, given the same code is being called to generate the _expected_ result and the _actual_ result. How could they not match?

An option would be to include in the test case a raw JSON payload and a structure containing the expected values we should see in the generated exception. This would make it clearer what is being tested, and give us more confidence that the result is what we expect.

source/Octopus.Client.Tests/Exceptions/OctopusExceptionFactoryFixture.cs
```diff
@@ -0,0 +1,304 @@
...
+                httpResponseMessage.Content = new StringContent(content);
+                resultValidationObject = OctopusExceptionFactory.OctopusErrorsContractFromBody(content);
+            }
+
+            var createdException =
+                await OctopusExceptionFactory.CreateException(httpResponseMessage) as OctopusValidationException;
+
+            using (new AssertionScope())
+            {
+                createdException.Should().NotBeNull();
```
adam-mccoy, line 51: Minor nitpick, but this assertion needn't be inside the scope, given if it fails there's no point continuing with the others.

source/Octopus.Client.Tests/Exceptions/OctopusExceptionFactoryFixture.cs
michaelongso: As discussed, the dependency on the factory is not the right approach as it will not provide safety net when there is a change in the factory. I've updated the implementation with `string payload, expected exception` pattern.

source/Octopus.Server.Client/Exceptions/OctopusExceptionFactory.cs
adam-mccoy: This can be made private again.

source/Octopus.Client.Tests/Exceptions/OctopusExceptionFactoryFixture.cs
```diff
@@ -0,0 +1,321 @@
...
+            var errorsValue = new[] { "Additional Errors" };
+            var jObject = new JObject
+            {
+                { "ErrorMessage", errorMessageValue },
+                { "Errors", JArray.FromObject(errorsValue) },
+                { "HelpText", helpTextValue },
+                { "Random", "not relevant"}
+            };
+
+            yield return new TestCaseData(jObject.ToString(), new OctopusValidationException(httpStatusCode, errorMessageValue, errorsValue) { HelpText = helpTextValue });
```
adam-mccoy, line 69: It's a clever way to generate the test data, but I think it'd be better to be actually _see_ the JSON. I can guess pretty well what the result of `jObject.ToString()` would be, but I'm not exactly sure. Might just be my personal preference, so I won't hold the PR up with it.

adam-mccoy (approved):
Looks good, thanks for that 👍 
I've left a couple of things to consider.

source/Octopus.Client.Tests/Exceptions/OctopusExceptionFactoryFixture.cs
michaelongso: I prefer this because injecting const variable on JSON string can get a bit messy. It's an eye sore as well with the amount of quotes it needs.

PR #720, opened 2022-12-08
Fix multipart/form-data file uploads for sync client

shaunmarx:
As part of migrating some of our nancy endpoints to MVC in asp.net core, we noticed that file uploads were crashing without any additional information to help us. We tracked this down to our client not including the trailing boundary `--` at the end of our form data content.

Since our sync client uses the old HttpWebRequest class, where we manually construct these boundaries, the content we were sending didn't quite match the content from our `async` client. Although not exactly the same, these are much more aligned now making it possible to use the same checks against the request in our tests for both types of clients.


andrewabest (approved):
There is a note about the async test that isn't awaiting anything. Fix looks fine given you've verified the behaviour / compatibility.

shaunmarx:
> There is a note about the async test that isn't awaiting anything. Fix looks fine given you've verified the behaviour / compatibility.
Ah good catch, I have updated the test :)

PR #721, opened 2022-12-09
Fix multipart trailer for sync client

shaunmarx:
We previously updated our client to include a [trailing boundary](https://github.com/OctopusDeploy/OctopusClients/pull/720), but unfortunately during a rebase it was removed by accident. This PR fixes this oversight and adds the appropriate trailer boundary back into multipart data data being sent via the sync client.


MissedTheMark (approved):
Pulled this branch and built client.  Tested against the update space logo endpoint and it worked fine 👍

PR #723, opened 2022-12-19
Convert to us GHA matrix

stevencl840:
In order to simplify the code used to test Octopus Clients against a variety of Linux distributions, it has been converted to use Matrix in the build process.

Significantly reduces the size of the workflow.


skermajo (approved):
Thanks - that's a lot easier to maintain now 👍

PR #724, opened 2022-12-21
Fix Rocky Linux 9 build and test

stevencl840:
Nightly build had started failing. A check of the logs revealed that gzip was missing. It was not missing before, but I guess this is why we do nightly builds. I have added gzip to the rocky9 test process and all seems to work correctly now.


skermajo (approved):
Nice 👍

PR #726, opened 2023-01-10
Add ForcePackageDownload to Runbook, Project & ProjectSetting resources

mik-ky:
Adding a ForcePackageDownload field for Runbooks & Project settings.

Related Server changes: https://github.com/OctopusDeploy/OctopusDeploy/pull/15914.

sc-30499

### DIFF
--- a/source/Octopus.Server.Client/Model/ProjectResource.cs
+++ b/source/Octopus.Server.Client/Model/ProjectResource.cs
@@ -125,6 +125,10 @@ namespace Octopus.Client.Model
         [Writeable]
         [Obsolete("Use " + nameof(DeploymentSettingsResource) + " instead on the `deploymentsettings` API.")]
         public string DeploymentChangesTemplate { get; set; }
+
+        [Writeable]
+        [Obsolete("Use " + nameof(DeploymentSettingsResource) + " instead on the `deploymentsettings` API.")]
+        public bool ForcePackageDownload { get; set; }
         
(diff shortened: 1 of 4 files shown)


source/Octopus.Server.Client/Model/ProjectResource.cs
```diff
@@ -125,6 +125,10 @@ public ProjectResource(string id, string name, string slug) : this()
         [Writeable]
         [Obsolete("Use " + nameof(DeploymentSettingsResource) + " instead on the `deploymentsettings` API.")]
         public string DeploymentChangesTemplate { get; set; }
+
+        [Writeable]
+        [Obsolete("Use " + nameof(DeploymentSettingsResource) + " instead on the `deploymentsettings` API.")]
```
IsaacCalligeros95, line 130: Should this Obsolete attribute be removed?

IsaacCalligeros95 (changes requested):
Minor comment on the Obsolete attribute, other than that LGTM

source/Octopus.Server.Client/Model/ProjectResource.cs
mik-ky: Nope, I don't think so. I think that Obsolete is there to guide the users to use the newer deploymentSettings API (where, this new flag belongs) rather than directly via the project resource. We'd still need to have this on the ProjectResource itself.

IsaacCalligeros95 (approved):

PR #727, opened 2023-01-17
Add OctopusClient.Post with no template parameters

rain-on:
Currently all calls to OctopusClient.Post require the type being transmitted to be supplied as a Template parameter, which means anonymous types canno be posted.

This change adds two new Post methods which do not require the type being transmitted, and will instead rely on the newtonsoft serialisation to ensure the body of the request is correctly serialised.


borland:
@rain-on do you still want to proceed with this PR? or abandon it in favour of the less-impactful `Post<object>(...)`

PR #736, opened 2023-03-23
Add cancellation tokens for IDeploymentSettingsRepository

matt-richardson:
To make them more consistent with other repositories.


borland (approved):

stevencl840 (approved):
Nice move in the right direction

stevencl840:
Is the framework approved file not affected?

matt-richardson:
> Is the framework approved file not affected?

I dont think so... I think that only allows the sync client.

PR #742, opened 2023-05-04
Update push-package and setup-dotnet actions

desruc:
# Background

[sc-45166]

A nightly build didn't show up in deploy for OctopusClients. This is because there was a network timeout while pushing the nuget packages to octopus. The current version of the `push-packages-action` (v1) installs the .NET CLI and uses it to push packages. It was this installation that timed out.

This PR updates `push-packages-action` to v3 as it doesn't need the CLI at all.

It also updates the `setup-dotnet` action to v3 to stop the `Node.js 12 actions are deprecated.` warnings in the action logs.

We will still see the `The set-output command is deprecated and will be disabled soon` warning for now. We need swap over to `Octopus.OctoVersion.Tool` from `OctoVersion.Tool` to fix these warnings but that is blocked until https://github.com/nuke-build/nuke/pull/1104 is released in Nuke.

# Results

- Update `push-package-action` to `v3`
- Update `setup-dotnet` to `v3`


borland (approved):
LGTM!

PR #746, opened 2023-05-10
Isaac/retention policy cron expression

IsaacCalligeros95:
Adding Retention Policy Configuration with Cron Expression. See https://github.com/OctopusDeploy/OctopusDeploy/pull/18021 for details.

[sc-40132]

### DIFF
--- a/source/Octopus.Server.Client/Model/RetentionPoliciesConfigurationResource.cs
+++ b/source/Octopus.Server.Client/Model/RetentionPoliciesConfigurationResource.cs
@@ -3,18 +3,14 @@ using Octopus.Client.Extensibility.Attributes;
 
 namespace Octopus.Client.Model
 {
-    public class RetentionPoliciesConfigurationResource : IResource
+    public class RetentionPoliciesConfigurationResource : Resource
     {
         public RetentionPoliciesConfigurationResource()
         {
             Id = "retentionpolicies";
         }
 
-        public string Id { get; set; }
-
         [Writeable]
         public string CronExpression { get; set; } = "0 */4 * * *";
-
-        public LinkCollection Links { get; set; }
     }
 }
--- a/source/Octopus.Server.Client/Model/RetentionPoliciesConfigurationResource.cs
+++ b/source/Octopus.Server.Client/Model/RetentionPoliciesConfigurationResource.cs
@@ -1,0 +1,19 @@
+using Octopus.Client.Extensibility;
+using Octopus.Client.Extensibility.Attributes;
+
+namespace Octopus.Client.Model
+{
+    public class RetentionPoliciesConfigurationResource : IResource
+    {
+        public RetentionPoliciesConfigurationResource()
+        {
+            Id = "retentionpolicies";
+        }
+
+        public string Id { get; set; }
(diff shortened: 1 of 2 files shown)


source/Octopus.Server.Client/Model/RetentionPoliciesConfigurationResource.cs
```diff
@@ -0,0 +1,20 @@
...
+namespace Octopus.Client.Model
+{
+    public class RetentionPoliciesConfigurationResource : IResource
+    {
+        public RetentionPoliciesConfigurationResource()
+        {
+            Id = "retentionpolicies";
+        }
+
+        public string Id { get; set; }
```
veochen-octopus, line 13: Maybe dumb question, why is the `Id` on a configuration resource? Does this mean we can set the id?

IsaacCalligeros95: I'm not exactly sure but we store Configuration documents as ID, json pairs.
image
These are all defined in server with the base class ConfigurationDocument which takes the ID as a part of the constructor. I assume the Id is just there for matching purposes and can't or at least shouldn't be changed.

benPearce1: This should probably inherit from `Resource` which already contains an Id property.
If I remember correctly there should be no POST to create configuration records, so you can't just set any Id and POST it to the API, should require a GET and PUT.

IsaacCalligeros95: > This should probably inherit from `Resource` which already contains an Id property. If I remember correctly there should be no POST to create configuration records, so you can't just set any Id and POST it to the API, should require a GET and PUT.

The implementations are a bit all over the shop, some like ArchiveLimitConfigurationResource and EventRetentionConfigurationResource use IResource. These we're the most recent implementations of ConfigurationResources so I went with that approach. I've updated to use Resource and left the others as is, let me know you think it's worth updating those too.

benPearce1: If you change them, does it change the surface area test?

IsaacCalligeros95: Yep, 
image

benPearce1: Then just leave it as is

benPearce1 (approved):

source/Octopus.Server.Client/Model/RetentionPoliciesConfigurationResource.cs
veochen-octopus: Only reason I ask is that configuration's ID is a db level implementation detail and is used to map between a db row and its correct configuration type, i.e. the same id always maps to the same type and vice verse. Apart from the document store nothing else should need to know about the id. So IMO exposing the id in the resource is leaky. But seems like we don't have any strict rules around this. Just raising for future consideration.

PR #748, opened 2023-06-08
add oci registry feedtype enum and resource

hnrkndrssn:
This PR adds the new OCI Registry feed type as we're adding support for OCI Registries in Octopus Server.

[sc-43537]


IsaacCalligeros95 (approved):
LGTM

PR #751, opened 2023-07-18
Find by partial name support

ankithkonda:
Add the ability to find by partial name. Server API has supported this for a while now, just updating client to make use of it.

First one to support this is tenants

Internal discussion with PET team - https://octopusdeploy.slack.com/archives/C04LWDDAC6L/p1689661870773089

### DIFF
--- a/source/Octopus.Server.Client/Repositories/Async/BasicRepository.cs
+++ b/source/Octopus.Server.Client/Repositories/Async/BasicRepository.cs
@@ -6,6 +6,7 @@ using System.Threading;
 using System.Threading.Tasks;
 using Octopus.Client.Exceptions;
 using Octopus.Client.Extensibility;
+using Octopus.Client.Extensions;
 using Octopus.Client.Model;
 using Octopus.Client.Util;
 using Octopus.Client.Validation;
@@ -264,7 +265,7 @@ namespace Octopus.Client.Repositories.Async
             return FindMany(r =>
             {
                 var named = r as INamedResource;
-                return named != null && named.Name.Contains(partialName);
+                return named != null && named.Name.Contains(partialName, StringComparison.OrdinalIgnoreCase);
             }, path, pathParameters);
         }
         
--- a/source/Octopus.Server.Client/Repositories/Async/BasicRepository.cs
+++ b/source/Octopus.Server.Client/Repositories/Async/BasicRepository.cs
@@ -254,15 +254,15 @@ namespace Octopus.Client.Repositories.Async
             return await Client.Get<List<TResource>>(link, parameters, cancellationToken).ConfigureAwait(false);
         }
 
-        public Task<List<TResource>> FindByPartialName(string partialName, string path, object pathParameters, CancellationToken cancellationToken)
+        public async Task<List<TResource>> FindByPartialName(string partialName, string path, object pathParameters, CancellationToken cancellationToken)
         {
-            ThrowIfServerVersionIsNotCompatible(cancellationToken).ConfigureAwait(false);
+            await ThrowIfServerVersionIsNotCompatible(cancellationToken);
 
             partialName = (partialName ?? string.Empty).Trim();
             if (pathParameters == null)
                 pathParameters = new { partialName = partialName};
 
-            return FindMany(r =>
+            return await FindMany(r =>
             {
                 var named = r as INamedResource;
                 return named != null && named.Name.Contains(partialName, StringComparison.OrdinalIgnoreCase);
--- a/source/Octopus.Server.Client/Repositories/Async/BasicRepository.cs
+++ b/source/Octopus.Server.Client/Repositories/Async/BasicRepository.cs
@@ -253,6 +253,26 @@ namespace Octopus.Client.Repositories.Async
             return await Client.Get<List<TResource>>(link, parameters, cancellationToken).ConfigureAwait(false);
         }
 
+        public Task<List<TResource>> FindByPartialName(string partialName, string path, object pathParameters, CancellationToken cancellationToken)
+        {
+            ThrowIfServerVersionIsNotCompatible(cancellationToken).ConfigureAwait(false);
+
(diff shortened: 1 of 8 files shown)


source/Octopus.Server.Client/Repositories/Async/BasicRepository.cs
```diff
@@ -253,6 +253,26 @@ public async Task<List<TResource>> GetAll(CancellationToken cancellationToken)
...
+
+            partialName = (partialName ?? string.Empty).Trim();
+            if (pathParameters == null)
+                pathParameters = new { partialName = partialName};
+
+            return FindMany(r =>
+            {
+                var named = r as INamedResource;
+                return named != null && named.Name.Contains(partialName);
+            }, path, pathParameters);
```
rosslovas, line 268: It looks like similar methods like `FindByName` are case insensitive, but this is case sensitive. It'd be good to keep them the same. I think you'll need to `IndexOf` because we're on `netstandard2.0`, the case insensitive `Contains` overload wasn't added til later IIRC:

```suggestion
            return FindMany(
                r => r is INamedResource named && named.Name.IndexOf(partialName, StringComparison.OrdinalIgnoreCase) != -1,
                path,
                pathParameters);
```

Might be worth double checking that the tenants endpoint is case insensitive too.

source/Octopus.Server.Client/Repositories/Async/BasicRepository.cs
```diff
@@ -253,6 +253,26 @@ public async Task<List<TResource>> GetAll(CancellationToken cancellationToken)
             return await Client.Get<List<TResource>>(link, parameters, cancellationToken).ConfigureAwait(false);
         }
 
+        public Task<List<TResource>> FindByPartialName(string partialName, string path, object pathParameters, CancellationToken cancellationToken)
+        {
+            ThrowIfServerVersionIsNotCompatible(cancellationToken).ConfigureAwait(false);
+
+            partialName = (partialName ?? string.Empty).Trim();
+            if (pathParameters == null)
+                pathParameters = new { partialName = partialName};
```
rosslovas, line 262: Minor nitpick: could use `??=` here

source/Octopus.Server.Client/Repositories/Async/BasicRepository.cs
ankithkonda: Thanks Ross! it looks like we do use .contains in other places

https://github.com/OctopusDeploy/OctopusClients/blob/b883446b1286904f102ca46013eee4927f02a967/source/Octopus.Server.Client/Repositories/Async/AccountRepository.cs#L76

https://github.com/OctopusDeploy/OctopusClients/blob/0a9f0eb5ecf9a9b064135ae4233bdd57544cfb5b/source/Octopus.Server.Client/OctopusAsyncRepository.cs#L218

Would IndexOf still be preferred? 

Good catch on the case sensitivity, I'll take a look

ankithkonda: Actually all good, was easy enough to add in an extension where you can pass in the OrdinalIgnoreCase

source/Octopus.Server.Client/Repositories/Async/BasicRepository.cs
ankithkonda: I might keep this as-is for consistency. Also subjectively i find this easier to read :)

source/Octopus.Server.Client/Repositories/Async/BasicRepository.cs
rosslovas: `Contains` is preferred where it's possible, it's just that `Contains(string, StringComparison)` wasn't added til .NET Standard 2.1 so it's the only option unless we upgrade. Your extension method is a solid approach 😄

source/Octopus.Server.Client/Repositories/Async/BasicRepository.cs
```diff
@@ -253,6 +254,26 @@ public async Task<List<TResource>> GetAll(CancellationToken cancellationToken)
             return await Client.Get<List<TResource>>(link, parameters, cancellationToken).ConfigureAwait(false);
         }
 
+        public Task<List<TResource>> FindByPartialName(string partialName, string path, object pathParameters, CancellationToken cancellationToken)
+        {
+            ThrowIfServerVersionIsNotCompatible(cancellationToken).ConfigureAwait(false);
```
rosslovas, line 259: It just occurred to me that we need to `await` this.

Looks like a bug that we're not awaiting this in a bunch of other methods in this class too 😞 You don't have to fix them now (besides this one) unless you want to, if you don't PET will handle it.

ankithkonda: Yeah no worries, I'll fix up my usage of it. Fixing the others is probably best to leave for a standalone PR

source/Octopus.Server.Client/Repositories/Async/BasicRepository.cs
```diff
@@ -253,6 +254,26 @@ public async Task<List<TResource>> GetAll(CancellationToken cancellationToken)
             return await Client.Get<List<TResource>>(link, parameters, cancellationToken).ConfigureAwait(false);
         }
 
+        public async Task<List<TResource>> FindByPartialName(string partialName, string path, object pathParameters, CancellationToken cancellationToken)
+        {
+            await ThrowIfServerVersionIsNotCompatible(cancellationToken);
```
rosslovas, line 259: We need to keep the `.ConfigureAwait(false)` as this is library code and it could potentially affect consumers using custom synchronisation contexts:

```suggestion
            await ThrowIfServerVersionIsNotCompatible(cancellationToken).ConfigureAwait(false);
```

ankithkonda: Huh interesting TIL, Updated it :)

rosslovas (approved):
A solid improvement here, nice one 🎉

PR #755, opened 2023-08-03
Multiple Git dependencies support

MissedTheMark:


source/Octopus.Server.Client/Model/GitDependencyResource.cs
```diff
@@ -1,21 +1,27 @@
...
+
 namespace Octopus.Client.Model;
 
 /// <summary>
 /// Represents a reference from a deployment-process' action to a Git resource.
 /// </summary>
 public class GitDependencyResource
 {
-    public GitDependencyResource(string repositoryUri, string defaultBranch, string gitCredentialType, string? gitCredentialId = null)
+    public GitDependencyResource(string repositoryUri, string defaultBranch, string gitCredentialType, string? gitCredentialId = null, string? filePathFilters = null, string? name = null)
```
johnsimons, line 11: Wondering if it would be better to make `filePathFilters`  default to `[]` instead of nullable.
Actually shouldn't `filePathFilters` be an array?

johnsimons (changes requested):

source/Octopus.Server.Client/Model/GitDependencyResource.cs
MissedTheMark: The string is split into an array at the back-end, which thinking about it now is a bit clunky from a client perspective 🤔

MissedTheMark: Lets leave it as is for now (I need these changes in for our bug bash), and I'll look into this today.

johnsimons (approved):
LGTM

zentron (approved):

PR #756, opened 2023-08-04
Fix net462 tests and some async improvements

borland:
# Background

We noticed that some of the async/await code in the async `BasicRepository` was inconsistent.
As part of routine maintenance, I cleaned this up.

In doing so, I discovered that the tests had not been running on the `net462` target for quite some time, due to [a bug in the NUnit test runner](https://github.com/nunit/nunit-console/issues/1176) that occurred when the .NET 7 SDK was present on the host machine. 
This issue was fixed in September 2022, and upon upgrading the NUnit runner I discovered that a few breakages/incompatibilities had crept into the codebase.

# Results

- Enables Microsoft analyzers `CS4014`, `CA1849`, `CA2012`, `CA2016` which help ensure we are handling async/await correctly.
- Installs the [AsyncFixer](https://github.com/semihokur/AsyncFixer) third-party analyzer for additional safety nets.
- Updates `LangVersion` to 8 for `Octopus.Client.csproj`
   - Note: This should have no effect at runtime for people who simply consume the binary DLL via a Nuget package, it's just at the compiler level
- Tweaks a couple of places in the code to handle platform differences between .NET Framework and .NET 6
- Consistently applies `ConfigureAwait(false)` and `async` methods in `BasicRepository.cs`, `MixedScopeBaseRepository.cs` and `UserInvitesRepository`


source/Octopus.Client.E2ETests/AssemblyExtensions.cs
```diff
@@ -1,19 +1,22 @@
...
     internal static class AssemblyExtensions
     {
         public static string FullLocalPath(this Assembly assembly)
         {
-            var codeBase = assembly.CodeBase;
+#if NETFRAMEWORK
+            var codeBase = assembly.CodeBase ?? throw new NotSupportedException($"Cannot get codebase for assembly {assembly}");
+#else
+            var codeBase = assembly.Location;
+#endif
```
adam-mccoy, line 15: Could we just use `Assembly.Location` for both frameworks?

source/Octopus.Client.E2ETests/Octopus.Client.E2ETests.csproj
```diff
@@ -25,9 +25,10 @@
             <IncludeAssets>runtime; build; native; contentfiles; analyzers</IncludeAssets>
         </PackageReference>
         <PackageReference Include="Microsoft.NET.Test.Sdk" Version="15.3.0" />
-        <PackageReference Include="NSubstitute" Version="4.3.0" />
-        <PackageReference Include="NUnit" Version="3.8.1" />
-        <PackageReference Include="NUnit3TestAdapter" Version="3.8.0" />
+        <PackageReference Include="NSubstitute" Version="4.4.0" />
+        <PackageReference Include="NUnit" Version="3.13.3" />
+        <PackageReference Include="NUnit3TestAdapter" Version="4.5.0" />
+        <PackageReference Include="System.IO.Compression" Version="4.3.0" />
```
adam-mccoy, line 31: Was the dependency on `System.IO.Compression` added intentionally?

adam-mccoy (approved):
LGTM. Just a couple of questions.

source/Octopus.Client.E2ETests/AssemblyExtensions.cs
borland: We *most likely* can.

[Microsoft's documentation](https://learn.microsoft.com/en-us/dotnet/api/system.reflection.assembly.location?view=net-7.0) contains this:

> **Remarks** 
In .NET 5 and later versions, for bundled assemblies, the value returned is an empty string.
>
> .NET Framework only: If the loaded file was [shadow-copied](https://learn.microsoft.com/en-us/dotnet/framework/app-domains/shadow-copy-assemblies), the location is that of the file after being shadow-copied. To get the location before the file has been shadow-copied, use the [CodeBase](https://learn.microsoft.com/en-us/dotnet/api/system.reflection.assembly.codebase?view=net-7.0) property.

Out of caution, I chose to leave the code untouched for NETFRAMEWORK.

source/Octopus.Client.E2ETests/Octopus.Client.E2ETests.csproj
borland: Yes, the newer version of `NUnit3TestAdapter` takes a dependency on it, and it's not built into the BCL in NETFRAMEWORK

PR #759, opened 2023-08-28
Add releaseHistoryCount to GET project progression

skermajo:
This is a partner PR to https://github.com/OctopusDeploy/OctopusDeploy/pull/19881, which added `releaseHistoryCount` to the project progression link template.

This PR adds the optional `releaseHistoryCount` value when sending a [GetProjectProgressionRequest](https://github.com/OctopusDeploy/OctopusDeploy/blob/2be9c86e473100bf68e83b8bbf3628aa7e03bb38/source/Octopus.Core/Features/Progression/MessageContracts/GetProjectProgressionRequest.cs) from the `IProjectRepostitory`.

[sc-58183]

### DIFF
--- a/source/Octopus.Server.Client/Repositories/Async/ProjectRepository.cs
+++ b/source/Octopus.Server.Client/Repositories/Async/ProjectRepository.cs
@@ -61,7 +61,7 @@ namespace Octopus.Client.Repositories.Async
         
         [Obsolete("Please use the overload with cancellation token instead.", false)]
         Task<ProgressionResource> GetProgression(ProjectResource project);
-        Task<ProgressionResource> GetProgression(ProjectResource project, CancellationToken cancellationToken);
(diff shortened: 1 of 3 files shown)


source/Octopus.Server.Client/Repositories/Async/ProjectRepository.cs
```diff
@@ -268,9 +268,10 @@ public Task<ProgressionResource> GetProgression(ProjectResource project)
             return GetProgression(project, CancellationToken.None);
         }
 
-        public Task<ProgressionResource> GetProgression(ProjectResource project, CancellationToken cancellationToken)
+        public Task<ProgressionResource> GetProgression(ProjectResource project, CancellationToken cancellationToken, int? releaseHistoryCount = null)
```
rosslovas, line 271: Even though the overload without a cancellation token is marked `[Obsolete]`, we should keep parity with the other overload by adding the `releaseHistoryCount` parameter there too, as it's specifically the lack of a cancellation token that makes it obsolete.

rosslovas (approved):

PR #762, opened 2023-08-29
Azure OIDC

veochen-octopus:
Adding Azure OIDC support. Need to think about merge timing.

[sc-57221]


hnrkndrssn:
> Need to think about merge timing.

The builds should be fixed now so we can reference pre-release versions of `Octopus.Server.Client` in Octopus and still get green builds.

IsaacCalligeros95 (approved):
LGTM

PR #763, opened 2023-08-31
Change defaults for RPC Retries

nathanwoctopusdeploy:
Change PRC retries to be on by default and set the retry duration to 2.5 minutes for new machine policies


sburmanoctopus (approved):
LGTM

PR #766, opened 2023-09-25
Add username password account variables and control type

susanpann:
[sc-58725]
Related https://github.com/OctopusDeploy/OctopusDeploy/pull/20170

### DIFF
--- a/source/Octopus.Server.Client/Model/VariableType.cs
+++ b/source/Octopus.Server.Client/Model/VariableType.cs
@@ -8,6 +8,6 @@
         AmazonWebServicesAccount,
         AzureAccount,
         WorkerPool,
-        UsernamePassword
+        UsernamePasswordAccount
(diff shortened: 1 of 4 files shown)


source/Octopus.Server.Client/Model/VariableType.cs
```diff
@@ -7,6 +7,7 @@ public enum VariableType
         Certificate,
         AmazonWebServicesAccount,
         AzureAccount,
-        WorkerPool
+        WorkerPool,
+        UsernamePassword
```
tothegills, line 11: This type is a slightly different convention to the other accounts, which all have the `Account` suffix.

tothegills (commented):
`UsernamePasswordAccount` would better match the existing naming convention.

tothegills (approved):

PR #770, opened 2023-10-03
Add support for token authentication

scme0:
Adding the ability for OctopusClient to use Tokens for authentication.

This does not include the ability for OctopusClient to facilitate a OIDC Token Exchange, and just allows the client to use a valid JWT Token that it is given.

This is being added so that Tentacle can use a Token to register itself as part of work over in [#project-k8s-agent](https://octopusdeploy.slack.com/archives/C05KK1G85J5). We want to be able to generate a token on server as part of generating a command for the user to run to install the Tentacle on their kubernetes cluster. This means for our use-case, Tentacle does not need to generate its own token and this update can be used instead.

[sc-60667]

### DIFF
--- a/source/Octopus.Client.Tests/PublicSurfaceAreaFixture.ThePublicSurfaceAreaShouldNotRegress..NETCore.approved.txt
+++ b/source/Octopus.Client.Tests/PublicSurfaceAreaFixture.ThePublicSurfaceAreaShouldNotRegress..NETCore.approved.txt
@@ -3,7 +3,7 @@ Octopus.Client
   class DefaultLinkResolver
     Octopus.Client.ILinkResolver
   {
-    .ctor(Uri, String)
+    .ctor(Uri, String = "/api")
     Boolean IsUsingSecureConnection { get; }
     Uri Resolve(String)
     String ToString()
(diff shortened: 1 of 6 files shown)


source/Octopus.Client.Tests/PublicSurfaceAreaFixture.ThePublicSurfaceAreaShouldNotRegress..NETCore.approved.txt
```diff
@@ -3,7 +3,7 @@ Octopus.Client
   class DefaultLinkResolver
     Octopus.Client.ILinkResolver
   {
-    .ctor(Uri, String)
+    .ctor(Uri, String = "/api")
```
scme0, line 6: I thought we could improve on this assent test by generating which properties are optional. I've only added optional properties so technically I didn't regress the public surface area but it's not clear from the required change to this test.

source/Octopus.Client.Tests/PublicSurfaceAreaFixture.ThePublicSurfaceAreaShouldNotRegress..NETCore.approved.txt
```diff
@@ -605,14 +605,17 @@ Octopus.Client
...
+    .ctor(Octopus.Client.ILinkResolver, String, ICredentials, Boolean = False)
     String ApiKey { get; }
     ICredentials Credentials { get; }
     Boolean IsUsingSecureConnection { get; }
     Octopus.Client.ILinkResolver OctopusServer { get; }
     IWebProxy Proxy { get; set; }
+    String Token { get; }
     Octopus.Client.OctopusServerEndpoint AsUser(String)
+    static Octopus.Client.OctopusServerEndpoint CreateWithApiKey(String, String)
+    static Octopus.Client.OctopusServerEndpoint CreateWithToken(String, String)
```
scme0, line 618: These are the changes that I've added/updated.

source/Octopus.Client.Tests/PublicSurfaceAreaFixture.ThePublicSurfaceAreaShouldNotRegress..NETCore.approved.txt
scme0: I reverted this because it was a PITA to manage the changes...

source/Octopus.Server.Client/OctopusServerEndpoint.cs
```diff
@@ -8,9 +8,31 @@ namespace Octopus.Client
...
-        readonly ILinkResolver octopusServer;
-        readonly string apiKey;
-        readonly ICredentials credentials;
+        /// <summary>
+        /// Create an instance with a Token to authenticate.
+        /// </summary>
+        public static OctopusServerEndpoint CreateWithToken(string octopusServerAddress, string token, ICredentials credentials = null)
+        {
+            return new OctopusServerEndpoint(octopusServerAddress, new TokenValue(token), credentials);
+        }
```
scme0, line 17: I couldn't find a particularly "nice" way to add the ability to specify a token without changing the existing public constructors in some way. My initial idea was to add an optional `bool isApiKey = false` parameter to the end of the appropriate constructor but that felt a bit too ugly. I decided to go for some public factory methods which are explicit and then have a private constructor taking a Token so it's not part of the public API.

source/Octopus.Server.Client/OctopusServerEndpoint.cs
```diff
@@ -8,9 +8,31 @@ namespace Octopus.Client
     /// </summary>
     public class OctopusServerEndpoint
     {
-        readonly ILinkResolver octopusServer;
-        readonly string apiKey;
-        readonly ICredentials credentials;
+        /// <summary>
+        /// Create an instance with a Token to authenticate.
+        /// </summary>
+        public static OctopusServerEndpoint CreateWithToken(string octopusServerAddress, string token, ICredentials credentials = null)
```
APErebus, line 14: I have no issue with this but just wondering if this our standard approach (to have factory method)?

source/Octopus.Server.Client/OctopusServerEndpoint.cs
```diff
@@ -61,17 +83,8 @@ public OctopusServerEndpoint(string octopusServerAddress, string apiKey)
...
-            if (string.IsNullOrWhiteSpace(octopusServerAddress))
-                throw new ArgumentException("An Octopus Server URI was not specified.");
-
-            if (!Uri.TryCreate(octopusServerAddress, UriKind.Absolute, out var uri) || !uri.Scheme.StartsWith("http"))
-                throw new ArgumentException($"The Octopus Server URI '{octopusServerAddress}' is invalid. The URI should start with http:// or https:// and be a valid URI.");
-
-            octopusServer = new DefaultLinkResolver(new Uri(octopusServerAddress));
-            this.apiKey = apiKey;
-            this.credentials = credentials ?? CredentialCache.DefaultNetworkCredentials;
         }
```
scme0, line 88: I just realised that this original constructor never tests if the `apiKey` is set to a value or not (it could be null). This is because before, the constructor that only takes the `octopusServerAddress` just uses this constructor with null.

My little refactor enforces this particular constructor to have an apiKey (not null or Whitespace) - you would need to use the constructor above to not set an API Key. 

Should I undo this? Technically it's a breaking change 😞

source/Octopus.Server.Client/OctopusServerEndpoint.cs
scme0: Sorry I just posted this comment after I asked you to look at it so you probably didn't see it: https://github.com/OctopusDeploy/OctopusClients/pull/770#discussion_r1351341841

source/Octopus.Client.Tests/PublicSurfaceAreaFixture.ThePublicSurfaceAreaShouldNotRegress..NETCore.approved.txt
scme0: Reverted/changed/not relevant anymore :)

source/Octopus.Server.Client/OctopusServerEndpoint.cs
scme0: I've rolled it back so it keeps the weird semi-validation. (validation on the server address, but not on the api key)

source/Octopus.Server.Client/OctopusServerEndpoint.cs
scme0: The main reason this is an issue is because the Token and the ApiKey are both strings, so adding a new constructor for the token would clash with the existing constructor/s which take API keys.

source/Octopus.Server.Client/OctopusServerEndpoint.cs
```diff
@@ -8,9 +8,31 @@ namespace Octopus.Client
...
+        /// </summary>
+        public static OctopusServerEndpoint CreateWithApiKey(string octopusServerAddress, string apiKey, ICredentials credentials = null)
+        {
+            return new OctopusServerEndpoint(octopusServerAddress, apiKey, credentials);
+        }
+
+        private OctopusServerEndpoint(string octopusServerAddress, BearerTokenValue bearerTokenValue, ICredentials credentials)
+        {
+            if (string.IsNullOrWhiteSpace(bearerTokenValue.Value))
+                throw new ArgumentException("Token is required.");
```
APErebus, line 30:
```suggestion
                throw new ArgumentException("BearerToken is required.");
```

source/Octopus.Server.Client/OctopusServerEndpoint.cs
```diff
@@ -8,9 +8,31 @@ namespace Octopus.Client
     /// </summary>
     public class OctopusServerEndpoint
     {
-        readonly ILinkResolver octopusServer;
-        readonly string apiKey;
-        readonly ICredentials credentials;
+        /// <summary>
+        /// Create an instance with a Token to authenticate.
+        /// </summary>
+        public static OctopusServerEndpoint CreateWithToken(string octopusServerAddress, string bearerToken, ICredentials credentials = null)
```
APErebus, line 14:
```suggestion
        public static OctopusServerEndpoint CreateWithBearerToken(string octopusServerAddress, string bearerToken, ICredentials credentials = null)
```
and also the assent tests

APErebus (approved):
2 minor nits, but apart from that 👍

PR #771, opened 2023-10-04
Include license in nuget package

emilol:
So that it gets picked up on our [credits](https://octopus.com/docs/credits) page, which is currently showing undefined.


source/Octopus.Server.Client/Octopus.Server.Client.csproj
```diff
@@ -7,6 +7,8 @@
     <AssemblyName>Octopus.Server.Client</AssemblyName>
     <GenerateAssemblyTitleAttribute>false</GenerateAssemblyTitleAttribute>
     <Authors>Octopus Deploy</Authors>
+    <PackageProjectUrl>https://github.com/OctopusDeploy/Nevermore</PackageProjectUrl>
```
adam-mccoy, line 10: Should this point to OctopusClients rather than Nevermore?

emilol: 🙈 fixed!

borland (approved):

PR #774, opened 2023-10-05
Add github workflow to v11.3

scme0:


scme0:
We have decided to update the version of OctopusClient in Tentacle instead of this approach.

PR #776, opened 2023-10-09
Add new net48 build target

scme0:
To allow us to upgrade `OctopusClient` in `OctopusTentacle`, we need to support .Net Framework 4.8

Unfortunately when we try and use the .Net Framework 4.6.2 we get some versioning issues with `System.Net.Http`.

In testing, creating a .Net Framework 4.8 version in the Nuget package eliminated the issues.


APErebus (approved):
Awesome 🚀

tekerson (approved):
Thanks for adding the relevant tests. It looks like this has had a pretty thorough manual vetting too!

PR #779, opened 2023-10-16
Append Timestamp to version when building locally

scme0:
Making a similar change to [this Tentacle change](https://github.com/OctopusDeploy/OctopusTentacle/pull/630) so that when building the package locally it automatically appends a timestamp on the version. This makes iterating on the client a lot quicker.


APErebus (approved):
Excellent 😄

PR #781, opened 2023-10-16
Fix Dockerfiles by making dotnet install more reliable

scme0:
After much fiddling it seems that making it install `dotnet` using the insecure `http://` makes it a lot more reliable as we seem to consistently get SSL issues causing the curl command to fail.

I'm just not sure if this is acceptable?


scme0:
It all seems to be working again with `https` this morning 🤷‍♂️

PR #783, opened 2023-10-22
.NET 8 prep: Use async method instead

YuKitsune:
# Summary

[sc-62738]

To help minimise the effort required to upgrade to .NET 8, we're back-porting any .NET 6 compatible changes from our .NET 8 branches.


rosslovas (approved):
Nice, was there an analyser change or something that brought this up? I'm all for it 😄

YuKitsune:
Yeah, one of the new .NET 8 analysers is a bit more vocal about calling non-sync methods from async methods when an async variant is available.

PR #786, opened 2023-10-26
Update EndpointResources and add interfaces

scme0:
This is one of the first steps toward adding the new Kubernetes Agent Target. #project-k8s-agent

To make things a bit neater I've moved a the `Uri` property to the base class and I've added interfaces to match the interfaces defined in Octopus Server.


scme0:
I decided not to bother changing this for now. Maybe in Machines API V2

PR #787, opened 2023-10-26
Adding changes to support new Kubernetes Agent Target Type

scme0:
This work is part of #project-k8s-agent to add a Kubernetes Agent to Octopus.

This PR adds the `KubernetesAgentEndpointResource` and supporting types and interfaces.
It also adds the `KubernetesAgent` option to `CommunicationStyle` and it makes it possible to register a Kubernetes Agent via the Client.

I've also moved the URI property from the child class (`ListeningTentacleEndpointResource` & `PollingTentacleEndpointResource`) into the parent class (`TentacleEndpointResource`) as this matches how we model these items in Server and it makes things a little neater. I will do the same thing in Server when I update the EndpointResource classes over there.

[sc-63126]


source/Octopus.Server.Client/Operations/IRegisterMachineOperationBase.cs
```diff
@@ -53,6 +54,21 @@ public interface IRegisterMachineOperationBase
...
+
+        /// <summary>
+        /// The default Container that the Kubernetes Agent will use to execute Jobs
+        /// </summary>
+        public string DefaultJobExecutionContainer { get; set; }
+
+        /// <summary>
+        /// The default Container Feed that the Kubernetes Agent will use to retrieve Job Execution Containers.
+        /// </summary>
+        public string DefaultJobExecutionContainerFeed { get; set; }
```
APErebus, line 70: Are we able to leave these out until we know the structure we want for these? I can see a scenario where we don't include these defaults for an MVP, but do them after

source/Octopus.Server.Client/Model/Endpoints/AgentCommunicationStyleResource.cs
```diff
@@ -0,0 +1,14 @@
+using Octopus.TinyTypes;
+
+namespace Octopus.Client.Model.Endpoints;
+
+public class AgentCommunicationStyleResource : CaseInsensitiveStringTinyType
```
APErebus, line 5: To avoid confustion with the _other_ `CommunicationStyle` thing, should we maybe call this something slightly different (like `CommunicationMode` or somesuch)?

source/Octopus.Server.Client/Model/Endpoints/KubernetesAgentEndpointResource.cs
```diff
@@ -0,0 +1,20 @@
...
+
+    public KubernetesAgentEndpointResource(TentacleEndpointConfigurationResource tentacleEndpointConfiguration, DeploymentActionContainerResource defaultJobExecutionContainer)
+    {
+        TentacleEndpointConfiguration = tentacleEndpointConfiguration;
+        DefaultJobExecutionContainer = defaultJobExecutionContainer;
+    }
+
+    public TentacleEndpointConfigurationResource TentacleEndpointConfiguration { get; set; }
+
+    public DeploymentActionContainerResource DefaultJobExecutionContainer { get; set; }
```
APErebus, line 19: As mentioned, let's leave this out for now (just make things less complicated)

APErebus (changes requested):
Nice work! Let's just leave out the default container stuff from this first cut of the work

source/Octopus.Server.Client/Operations/IRegisterMachineOperationBase.cs
scme0: Ah yep fair enough, good call 👍

source/Octopus.Server.Client/Model/Endpoints/AgentCommunicationStyleResource.cs
scme0: I went with `AgentCommunicationMode/Resource`. Does that work for you?

APErebus: yep, perfect

APErebus (approved):

PR #788, opened 2023-10-27
Fix AgentCommunicationMode property name

scme0:
I accidentally renamed this property when I was using Rider's refactor.

Just naming back to what I wanted it to be.


scme0:
[sc-63126]

APErebus (approved):
Sorry I missed that 🤦

PR #789, opened 2023-10-27
Add "Build and test outcome" job

rosslovas:


borland:
> it would take listing 30+ different status checks as required, and keeping that up-to-date with any future changes.

My understanding is that for PR validation using TeamCity, we need to preconfigure the status check, then post to it via the API, so maintaining these is burdensome.

But for GitHub Actions, it's been my experience that these all happen for free. I.e. you might queue a build that results in 300 status checks, but GHA creates and checks them all itself and you don't need to do anything if checks are added/removed over time

In that scenario, does the "outcome" job buy us anything?

rosslovas:
> In that scenario, does the "outcome" job buy us anything?

They happen for free, that part's not a problem. The annoying part is having to maintain the list of which ones are required.

PR #801, opened 2023-11-15
Update fluent assertions version and fix breaking changes

scme0:
I'm adding some tests and I've noticed that FluentAssertions is quite behind. I want to use new features so I'm giving updating it a crack. Only changing tests to fix breaking changes in FluentAssertions as we're updating over a few major versions.

I'm thinking if the build goes green we should be good to go 👍

### DIFF
--- a/source/Octopus.Client.Tests/Serialization/CrossPlatformDateTimeZoneJsonConverterFixture.cs
+++ b/source/Octopus.Client.Tests/Serialization/CrossPlatformDateTimeZoneJsonConverterFixture.cs
@@ -39,7 +39,7 @@ namespace Octopus.Client.Tests.Serialization
 
             date
                 .Should()
-                .BeCloseTo(expected, 1000); // Within one second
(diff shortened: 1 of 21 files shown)


source/Octopus.Client.Tests/Serialization/CrossPlatformDateTimeZoneJsonConverterFixture.cs
```diff
@@ -39,7 +39,7 @@ public void ShouldDeserializeDateTimeOffset(string serializedDate, bool hasTimeZ
 
             date
                 .Should()
-                .BeCloseTo(expected, 1000); // Within one second
+                .BeCloseTo(expected, TimeSpan.FromSeconds(1)); // Within one second
```
YuKitsune, line 42: Nit: Could probably remove these comments now :)

APErebus (approved):
LGTM

YuKitsune (approved):
Pretty straightforward, thank you for taking the time to upgrade this 🙇

PR #802, opened 2023-11-15
Ensure k8s reregistration is correctly

scme0:
# Background

Part of #project-k8s-agent

The standard behaviour for the `register-with` command (which is for register machines, as opposed to workers) will automatically set the `--force` flag. This means that it will "blat" over the existing target in server with whatever has been configured.

The `register-k8s-cluster` command followed the same formula originally but it needs to be updated as this doesn't work with kubernetes very well. As the Kubernetes Tentacle will be running in a cluster (duh), the pod could be stopped and restarted at anytime. If we use the existing functionality, the container will "blat" over the settings in server every time it restarts for whatever reason.

However there is also a risk that if the originally configuration is lost in k8s (if it's using a volume mount, the drive config could be deleted, if it's using a configmap or secret, that could be deleted too). If this happens, Tentacle would create a new certificate, thumbprint and other information on restart which would break the connection to server unless we re-register the tentacle.

# Result

The new solution is that when the arg `--force` is not provided and there is already an existing target with the same name, it will only update the Endpoint information, not other information which is user configurable (such as environments, tenants, target roles). This safeguards us from anyone fiddling with configmaps and what not while also maintaining user configured settings.

### DIFF
--- a/source/Octopus.Client.Tests/Operations/RegisterKubernetesClusterOperationFixture.cs
+++ b/source/Octopus.Client.Tests/Operations/RegisterKubernetesClusterOperationFixture.cs
@@ -1,10 +1,12 @@
 using System;
 using System.Collections.Generic;
+using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
 using FluentAssertions;
 using NSubstitute;
 using NUnit.Framework;
+using Octopus.Client.Exceptions;
 using Octopus.Client.Extensibility;
 using Octopus.Client.Model;
 using Octopus.Client.Model.Endpoints;
@@ -22,6 +24,7 @@ namespace Octopus.Client.Tests.Operations
         ResourceCollection<EnvironmentResource> environments;
         ResourceCollection<MachineResource> machines;
         ResourceCollection<MachinePolicyResource> machinePolicies;
+        private ResourceCollection<ProxyResource> proxies;
 
         [SetUp]
         public void SetUp()
@@ -32,6 +35,7 @@ namespace Octopus.Client.Tests.Operations
             operation = new RegisterKubernetesClusterOperation(clientFactory);
             serverEndpoint = new OctopusServerEndpoint("http://octopus", "ABC123");
 
+            proxies = new ResourceCollection<ProxyResource>(new ProxyResource[0], LinkCollection.Self("/foo"));
             environments = new ResourceCollection<EnvironmentResource>(new EnvironmentResource[0], LinkCollection.Self("/foo"));
             machines = new ResourceCollection<MachineResource>(new MachineResource[0], LinkCollection.Self("/foo"));
             machinePolicies = new ResourceCollection<MachinePolicyResource>(new MachinePolicyResource[0], LinkCollection.Self("/foo"));
@@ -45,6 +49,7 @@ namespace Octopus.Client.Tests.Operations
                     .Add("MachinePolicies", "/api/machinepolicies")
                     .Add("CurrentUser", "/api/users/me")
                     .Add("SpaceHome", "/api/spaces")
+                    .Add("Proxies", "/api/proxies")
             };
 
             client.Get<RootResource>(Arg.Any<string>(), Arg.Any<CancellationToken>()).Returns(rootDocument);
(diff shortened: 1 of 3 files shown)


source/Octopus.Client.Tests/Operations/RegisterKubernetesClusterOperationFixture.cs
```diff
@@ -101,5 +108,147 @@ public async Task ShouldCreateNewPollingKubernetesTentacle()
 
             machineResources.Should().ContainSingle().Which.Endpoint.Should().BeOfType<KubernetesTentacleEndpointResource>().Which.Should().BeEquivalentTo(new KubernetesTentacleEndpointResource(new PollingTentacleEndpointConfigurationResource("ABCDEF", "poll://ckyhfyfkcbmzjl8sfgch/")));
         }
+
+        [Test]
+        public async Task ShouldOnlyUpdateEndpointConnectionConfigurationOnExistingMachine()
```
scme0, line 113: I ended up adding all these tests from the `RegisterMachineOperationFixture` just to ensure the functionality is maintained for `RegisterKubernetesClusterOperation`, even if someone wants to change `RegisterMachineOperation`.

source/Octopus.Client.Tests/Operations/RegisterKubernetesClusterOperationFixture.cs
```diff
@@ -101,5 +108,147 @@ public async Task ShouldCreateNewPollingKubernetesTentacle()
...
+                Id = "machines/84",
+                Name = "Mymachine",
+                EnvironmentIds = new ReferenceCollection("environments-1"),
+                Endpoint = new KubernetesTentacleEndpointResource(new ListeningTentacleEndpointConfigurationResource("ABCDEF", "https://my-new-machine.test.com:10930/") {ProxyId = "proxy-1"}),
+                Links = LinkCollection.Self("/machines/whatever/1")
+            }, cfg => cfg.RespectingRuntimeTypes());
+        }
+
+        [Test]
+        public async Task ShouldUpdateExistingMachineWhenForceIsEnabled()
```
scme0, line 152: I ended up adding all these tests from the `RegisterMachineOperationFixture` just to ensure the functionality is maintained for `RegisterKubernetesClusterOperation`, even if someone wants to change `RegisterMachineOperation`.

source/Octopus.Client.Tests/Operations/RegisterKubernetesClusterOperationFixture.cs
scme0, line 113: This test ensures the new functionality is working correctly.

source/Octopus.Server.Client/Operations/RegisterMachineOperation.cs
```diff
@@ -65,16 +63,26 @@ public RegisterMachineOperation(IOctopusClientFactory clientFactory) : base(clie
         /// </exception>
         public override void Execute(IOctopusSpaceRepository repository)
```
scme0, line 64: Side note: It would be great if we could remove these non-async methods. It's an annoying amount of duplication and I'm not sure if they are used anymore? They don't seem to be tested.

source/Octopus.Server.Client/Operations/RegisterMachineOperation.cs
```diff
@@ -130,12 +138,11 @@ MachineResource GetMachine(IOctopusSpaceRepository repository)
             try
             {
                 existing = repository.Machines.FindByName(MachineName);
-                if (!AllowOverwrite && existing?.Id != null)
-                    throw new InvalidRegistrationArgumentsException($"A machine named '{MachineName}' already exists on the Octopus Server in the target space. Use the 'force' parameter if you intended to update the existing machine.");
```
scme0, line 134: This exception message is different to the one below that I removed. Not sure why as they are the async/non-async versions of the same method. 🤔 

I've opted for the async version as it seems to be the one that is used.

source/Octopus.Server.Client/Operations/RegisterMachineOperation.cs
APErebus: Interesting. Might be worth asking more widely if these could be trashed

source/Octopus.Server.Client/Operations/RegisterMachineOperation.cs
```diff
@@ -147,17 +154,34 @@ MachineResource GetMachine(IOctopusSpaceRepository repository)
...
-            var tenants = GetTenants(repository).ConfigureAwait(false);
-            await ValidateTenantTags(repository).ConfigureAwait(false);
-            var proxy = GetProxy(repository).ConfigureAwait(false);
+            var machine = await GetMachine(repository).ConfigureAwait(false);
+            var proxy = await GetProxy(repository).ConfigureAwait(false);
 
-            var machine = await machineTask;
-            ApplyBaseChanges(machine, await machinePolicy, await proxy);
-            ApplyDeploymentTargetChanges(machine, await selectedEnvironments, await tenants);
+            if (machine.Id == null || AllowOverwrite)
```
APErebus, line 160: a null `Id` here represents a new machine resource correct?

APErebus (approved):
Excellent stuff mate. Like it!

source/Octopus.Server.Client/Operations/RegisterMachineOperation.cs
scme0: Yep correct. I've updated the code to make that a bit more obvious 👍

scme0:
[sc-60420]

PR #804, opened 2023-11-22
DeploymentsRepository - allow FindBy to filter by tenants

rain-on:
This adds a new "FindBy" function which allows deployments to be filtered by TenantId.


source/Octopus.Server.Client/Repositories/Async/DeploymentRepository.cs
```diff
@@ -40,6 +40,11 @@ public async Task<ResourceCollection<DeploymentResource>> FindBy(string[] projec
         {
             return await Client.List<DeploymentResource>(await Repository.Link("Deployments").ConfigureAwait(false), new { skip, take, projects = projects ?? new string[0], environments = environments ?? new string[0] }).ConfigureAwait(false);
         }
+        
+        public async Task<ResourceCollection<DeploymentResource>> FindBy(string[] projects, string[] environments, string[] tenants, int skip = 0, int? take = null)
```
adam-mccoy, line 44: I imagine this should also be added to the synchronous [DeploymentRepository](https://github.com/OctopusDeploy/OctopusClients/blob/master/source/Octopus.Server.Client/Repositories/DeploymentRepository.cs) for consistency.

PR #805, opened 2023-11-22
Use .NET 8 SDK

YuKitsune:
[sc-65325]

Depends on:
- [x] https://github.com/OctopusDeploy/OctopusClients/pull/807

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -38,11 +38,13 @@ jobs:
       - uses: actions/checkout@v4
         with:
           fetch-depth: 0 # all
-      - name: Setup .NET 6.0
+
+      # Use the .NET 8 SDK to target .NET 8
+      - name: Setup .NET 8.0
         uses: actions/setup-dotnet@v3
         with:
-          dotnet-version: 6.0.x
-      
+          dotnet-version: 8.0.x
+
       # Adjustment is done prior to Nuke build as OCTOVERSION information is included in the result package.
       - name: Append OCTOVERSION_CurrentBranch with -nightly (for scheduled)
         if: github.event_name == 'schedule'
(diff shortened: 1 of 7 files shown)


.github/workflows/main.yaml
```diff
@@ -120,7 +138,7 @@ jobs:
   test-linux-distributions:
     strategy:
       matrix:
-        os: [ubuntu16, ubuntu18, ubuntu20, debian-buster, fedora35, centos7, rhel9, rocky9, amazon2, opensuse-leap15, opensuse-tumbleweed ]
+        os: [ubuntu16, ubuntu18, ubuntu20, debian-buster, fedora35, rhel9, rocky9, amazon2, opensuse-leap15, opensuse-tumbleweed ]
```
YuKitsune, line 141: [.NET8 isn't supported on Centos](https://learn.microsoft.com/en-us/dotnet/core/install/linux-centos#supported-distributions).

source/Octopus.Client.Tests/Octopus.Client.Tests.csproj
```diff
@@ -12,18 +12,18 @@
     <LangVersion>8</LangVersion>
   </PropertyGroup>
   <PropertyGroup Condition="!$([MSBuild]::IsOSUnixLike())">
-    <TargetFrameworks>net462;net48;net6.0</TargetFrameworks>
+    <TargetFrameworks>net462;net48;net6.0;net8.0</TargetFrameworks>
   </PropertyGroup>
   <PropertyGroup Condition="$([MSBuild]::IsOSUnixLike())">
-    <TargetFramework>net6.0</TargetFramework>
+    <TargetFrameworks>net6.0;net8.0</TargetFrameworks>
   </PropertyGroup>
```
YuKitsune, line 19: @borland Do we still need to target .NET 6?

borland: Yeah, if the library targets net6 then the tests should too

.github/workflows/main.yaml
```diff
@@ -38,11 +38,13 @@ jobs:
       - uses: actions/checkout@v4
         with:
           fetch-depth: 0 # all
-      - name: Setup .NET 6.0
+
```
borland, line 41: Because the tests target both 6 and 8, we should have both a setup net6 and a setup net8 action here

borland (approved):

PR #809, opened 2023-12-07
Add GitHub credential type for Git persistence settings

dylanlerch:
Allows the use of the upcoming GitHub app connections as Config as Code credentials


geofflamrock (approved):
LGTM

PR #818, opened 2024-02-07
Modify deployments freezes

tothegills:


source/Octopus.Server.Client/Model/DeploymentFreezes/ModifyDeploymentFreezeCommand.cs
```diff
@@ -0,0 +1,23 @@
+﻿using System;
+using System.Collections.Generic;
+using System.ComponentModel.DataAnnotations;
+
+namespace Octopus.Client.Model.DeploymentFreezes;
+
+public class ModifyDeploymentFreezeCommand
+{
+    [Required] public string Id { get; set; }
```
domenicsim1, line 9: I guess we spicing it up on the client with the required being on the same line ay? 🌶️

domenicsim1 (approved):
Nice 😎

source/Octopus.Server.Client/Model/DeploymentFreezes/ModifyDeploymentFreezeCommand.cs
tothegills: I accidentally auto-formatted the permission file and ended up with 1000 changes, so spicy!

PR #827, opened 2024-02-21
Add FeedCategory to FeedFilterResource

EddyMoulton:
Update client for changes made to the API model: https://github.com/OctopusDeploy/OctopusDeploy/pull/22888

### DIFF
--- a/source/Octopus.Server.Client/Model/Triggers/FeedFilterResource.cs
+++ b/source/Octopus.Server.Client/Model/Triggers/FeedFilterResource.cs
@@ -9,4 +9,7 @@ public class FeedFilterResource : TriggerFilterResource
 
     [Writeable]
     public List<DeploymentActionPackageResource> Packages { get; set; } = new();
+
+    [Writeable]
+    public string FeedCategory { get; set; }
 }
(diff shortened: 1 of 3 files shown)


source/Octopus.Server.Client/Model/Triggers/FeedFilterResource.cs
```diff
@@ -9,4 +9,7 @@ public class FeedFilterResource : TriggerFilterResource
 
     [Writeable]
     public List<DeploymentActionPackageResource> Packages { get; set; } = new();
+
+    [Writeable]
+    public string FeedCategory { get; set; }
```
MissedTheMark, line 14: Double checking - does this need to have a default value of empty string or will that get set on the other end if it's missing? (the server resource does have the default value)

MissedTheMark (approved):
LGTM.  One thing I just wanted to double check.

source/Octopus.Server.Client/Model/Triggers/FeedFilterResource.cs
EddyMoulton: I don't think it needs it, but that's purely based off other resources that also do not set a default value.

PR #830, opened 2024-02-23
Add KubernetesAgentDetails and UpgradeLocked properties

scme0:
# Background

#project-k8s-agent

As part of adding server-driven upgrades for the kubernetes agent, we need to fetch additional information about the Agent Version, the k8s namespace that the agent runs in and then helm release name during health checks.

I'm also adding UpgradeInformation in rather than shoehorning in the old `TentacleVersionDetails` class.

These changes are technically breaking but the only thing that is using the Version/Upgrade status is front end, specifically for displaying the version and upgrading the agent. Considering upgrades don't work yet and the agent hasn't been released yet, this isn't a big deal. Registration isn't broken because none of the properties required for registration have been moved or deleted.

Flow on PRs:
Tentacle: https://github.com/OctopusDeploy/OctopusTentacle/pull/826 (it doesn't require any actual changes)
Server: https://github.com/OctopusDeploy/OctopusDeploy/pull/24061 (updating API Resources, Entity Models, and mappers)

[sc-67693]


source/Octopus.Server.Client/Model/Endpoints/ListeningKubernetesTentacleEndpointConfigurationResource.cs
```diff
@@ -0,0 +1,20 @@
+using Octopus.Client.Extensibility.Attributes;
+
+namespace Octopus.Client.Model.Endpoints
+{
+    public class ListeningKubernetesTentacleEndpointConfigurationResource : KubernetesTentacleEndpointConfigurationResource, IListeningTentacleEndpointResource
```
scme0, line 5: This is a rename from `ListeningTentacleEndpointConfigurationResource` to `ListeningKubernetesTentacleEndpointConfigurationResource`. It's a mouthful but I want to make it clear that this is used for the KubernetesTentacle, not just any random Tentacle.

source/Octopus.Server.Client/Model/Endpoints/PollingKubernetesTentacleEndpointConfigurationResource.cs
```diff
@@ -0,0 +1,15 @@
+namespace Octopus.Client.Model.Endpoints
+{
+    public class PollingKubernetesTentacleEndpointConfigurationResource : KubernetesTentacleEndpointConfigurationResource, IPollingTentacleEndpointResource
```
scme0, line 3: This is a rename for `PollingTentacleEndpointConfigurationReosurce` to PollingKubernetesTentacleEndpointConfigurationResource`.

source/Octopus.Server.Client/Model/Endpoints/KubernetesTentacleEndpointConfigurationResource.cs
```diff
@@ -3,13 +3,13 @@
 
 namespace Octopus.Client.Model.Endpoints
 {
-    public abstract class TentacleEndpointConfigurationResource : ITentacleEndpointResource
+    public abstract class KubernetesTentacleEndpointConfigurationResource : ITentacleEndpointResource
```
scme0, line 6: I've renamed this and child classes to include `Kubernetes` because now it's got info to do with the Kubernetes Agent so this class needs to make it clear that it is only for Kubernetes Tentacle.

flin-8 (approved):
LGTM

source/Octopus.Server.Client/Model/Endpoints/TentacleEndpointConfigurationResource.cs
```diff
@@ -27,9 +27,6 @@ protected TentacleEndpointConfigurationResource(string thumbprint, string uri)
         [Writeable]
         public string Uri { get; set; }
 
-        [Writeable]
-        public TentacleDetailsResource TentacleVersionDetails { get; set; }
```
scme0, line 31: I've removed this as it's no longer required for Kubernetes Tentacles. The information stored in this property is now stored in a combination of `KubernetesAgentDetails` and `UpgradeInformation` properties directly on the `KubernetesTentacleEndpointResource`.

source/Octopus.Server.Client/Model/Endpoints/UpgradeStatusResource.cs
```diff
@@ -0,0 +1,15 @@
+using Octopus.TinyTypes;
+
+namespace Octopus.Client.Model.Endpoints;
+
+public class UpgradeStatusResource : CaseInsensitiveStringTinyType
```
scme0, line 5: My understanding is that it's bad form to use an Enum on a public API so I've used a TinyType instead. Is this ok?

source/Octopus.Server.Client/Model/Endpoints/KubernetesAgentUpgradeInformationResource.cs
```diff
@@ -0,0 +1,8 @@
+namespace Octopus.Client.Model.Endpoints;
+
+public record KubernetesAgentUpgradeInformationResource
```
scme0, line 3: I don't love this name, can you think of any better ones?

source/Octopus.Server.Client/Model/Endpoints/ListeningKubernetesTentacleEndpointConfigurationResource.cs
scme0: I've removed the rename as it's not worth it.

source/Octopus.Server.Client/Model/Endpoints/PollingKubernetesTentacleEndpointConfigurationResource.cs
scme0: Reverted as not worth it.

source/Octopus.Server.Client/Model/Endpoints/KubernetesTentacleEndpointConfigurationResource.cs
scme0: Reverted as not worth it.

source/Octopus.Server.Client/Model/Endpoints/KubernetesAgentUpgradeInformationResource.cs
flin-8: What's this meant to store?

scme0: Basically just those two properties: `UpgradeStatus` which can be `NoUpgrades`, `UpgradeAvailable`, `UpgradeSuggested`, and `UpgradeRequired`. These values are based on how out-of-date the current version is to the suggested version. Eg: if a version is a major version or more behind, it's `UpgradeRequired`, if it's a minor version out it's `UpgradeSuggested` etc.
The other property is `UpgradeLocked`. If this is set by the user in the target connectivity page, the target won't be updated until it's unlocked.

scme0: I've moved the UpgradeLocked and UpgradeStatus properties as we discussed. I've applied the same changes to the Server PR and I think it works pretty well. 👍

scme0: It looks like we [got two thumbs up](https://octopusdeploy.slack.com/archives/C04LWDDAC6L/p1713340955931229) for this approach from #team-backend-foundations.

source/Octopus.Server.Client/Model/Endpoints/KubernetesAgentDetailsResource.cs
```diff
@@ -0,0 +1,34 @@
...
+            string @namespace)
+        {
+            Version = version;
+            TentacleVersion = tentacleVersion;
+            UpgradeStatus = upgradeStatus;
+            HelmReleaseName = helmReleaseName;
+            Namespace = @namespace;
+        }
+
+        public string Version { get; set; }
```
APErebus, line 24: Is this the Agent version? If so, `AgentVersion` might be a bit clearer?

source/Octopus.Server.Client/Model/Endpoints/KubernetesAgentDetailsResource.cs
```diff
@@ -0,0 +1,34 @@
...
+
+        public string TentacleVersion { get; set; }
+
+        public UpgradeStatus UpgradeStatus { get; set; }
+
+        public string HelmReleaseName { get; set; }
+
+        public string Namespace { get; set; }
```
APErebus, line 32: `KubernetesNamespace`? might be clearer?

APErebus (approved):
Couple of non-blocking questions

source/Octopus.Server.Client/Model/Endpoints/KubernetesAgentDetailsResource.cs
scme0: Yep that's fair 👍

source/Octopus.Server.Client/Model/Endpoints/KubernetesAgentDetailsResource.cs
scme0: Yep fair call 👍

source/Octopus.Server.Client/Model/Endpoints/KubernetesAgentDetailsResource.cs
scme0: Done

source/Octopus.Server.Client/Model/Endpoints/KubernetesAgentDetailsResource.cs
scme0: Done

PR #833, opened 2024-03-06
FeedRepository to allow fetching of mulitple versions

rain-on:
Currently the FeedRepository only fetches a single version  -this change adds a function which populates the includeMultipleVersions such that more than 1 is returned.


rain-on:
obsolete

PR #837, opened 2024-04-14
Add helper methods for getting and setting tenant variables

adam-mccoy:
# Background

Getting and setting tenant variable values using the client is difficult due to how they're accessed from the `TenantVariableResource`. It requires navigating 2 or 3 nested layers of dictionaries, with no indication of which keys are to be used at each level - you just have to know.

Also, the template being used is specified by its ID, which is not meaningful to end-users, rather than it's name. A lookup has to be manually performed each time to translate the name to the ID.

# Results

Added convenience methods to the `TenantVariablesResource` and `TenantVariableEditor` classes to simplify getting and setting values for tenant variables.

Example of usage in https://github.com/OctopusDeploy/OctopusDeploy/pull/24166

### DIFF
--- a/source/Octopus.Client.Tests/Model/TenantVariableResourceFixture.cs
+++ b/source/Octopus.Client.Tests/Model/TenantVariableResourceFixture.cs
@@ -9,6 +9,33 @@ namespace Octopus.Client.Tests.Model
     public class TenantVariableResourceFixture
     {
         [Test]
+        public void TryGettingProjectVariable_Works()
+        {
+            var project = new ProjectResource("Projects-1", "TestProject", "test-project");
+            var environment = new EnvironmentResource { Id = "Environments-1" };
+            var templateId = Guid.NewGuid().ToString();
+            var resource = new TenantVariableResourceBuilder()
+                .WithProjectEnvironment(project, environment)
+                .WithProjectTemplate(project, action => action
+                    .WithId(templateId)
+                    .WithName("ProjectVariable1"))
+                .Build();
+
+            resource.ProjectVariables[project.Id].Variables[environment.Id][templateId] = "SomeValue";
+
+            resource.TryGetProjectVariableValue(project, environment, "ProjectVariable1", out var result)
+                .Should()
+                .BeTrue();
+
+            result.Should().BeEquivalentTo(new PropertyValueResource("SomeValue"));
+
+            resource.TryGetProjectVariableValue(project, environment, "MissingTemplate", out var emptyResult)
+                .Should()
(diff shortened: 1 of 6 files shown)


source/Octopus.Server.Client/Model/TenantVariableResource.cs
```diff
@@ -14,6 +15,78 @@ public class TenantVariableResource : Resource, IHaveSpaceResource
         
         // Token to validate that no variables for the given tenant have changed since this TenantVariableResource was last requested.
         public string ConcurrencyToken { get; set; }
+
```
borland, line 18: Prior to this PR, the Resource types were simple bags of data that would be easy to pass around and perhaps mock. Adding the methods increases the API surface and seeing as though they all operate over the existing data structures and don't introduce any new data or encapsulation, I feel like they'd be better as Extension Methods on `TenantVariableResource`, etc.

Tests and percieved API surface would be unchanged, but it draws a bit of a line between the core thing and the convenience helpers.

source/Octopus.Client.Tests/Model/TenantVariableResourceFixture.cs
```diff
@@ -0,0 +1,254 @@
...
+                .WhoseValue.Should().BeEquivalentTo(new PropertyValueResource("SomeValue"));
+        }
+    }
+
+    public class TenantVariableResourceBuilder
+    {
+        private Dictionary<string, TenantVariableResource.Project> projects = new Dictionary<string, TenantVariableResource.Project>();
+        private Dictionary<string, TenantVariableResource.Library> library = new Dictionary<string, TenantVariableResource.Library>();
+
+        public TenantVariableResourceBuilder WithProjectEnvironment(ProjectResource projectResource, EnvironmentResource environmentResource)
```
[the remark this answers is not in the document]
adam-mccoy: It's idiomatic in the client to always pass full resources around. I was being consistent with that.

source/Octopus.Server.Client/Model/TenantVariableResource.cs
adam-mccoy: I was originally going to put these on the "editor" type for tenant variables. Looking at `TenantEditor` as an example, it just delegates to method calls on the resource itself. I was content with conforming to this existing pattern rather than trying to establish a new one.

Put simply, no new ground is being broken here.

borland (approved):
Thanks for answering my questions/comments. I've barely used Octopus.Client so my sense for "precedent" is off. Approved

PR #838, opened 2024-04-16
Rename TentacleEndpointConfigurationResource to make it clear it's only for KubernetesTentacles

scme0:
*TentacleEndpointConfigurationResource -> *KubernetesTentacleEndpointConfigurationResource

I want to make it clear these classes are for KubernetesTentacles only. I'll make a similar change in server after this merges.


scme0:
Closing because I don't think this is worth it.

PR #839, opened 2024-04-17
Change ProcessExternalFeedTriggers task name

EddyMoulton:
We'd like to rename a task from `PollFeedsForTriggers` to `ProcessExternalFeedTriggers`. I'll do this in two parts in `OctopusClients` to avoid any chance of a breaking change.

[sc-76099]


MissedTheMark (approved):
Tick.

PR #840, opened 2024-04-18
Remove PollFeedsForTriggers task

EddyMoulton:
Follow up to #839

Removes the now-unused `PollFeedsForTriggers` task

[sc-76099]


MissedTheMark (approved):
Nice.

PR #849, opened 2024-05-13
Add ProjectId to TaskResource

N-lson:
Relates to https://github.com/OctopusDeploy/OctopusDeploy/pull/24637


N-lson:
> Just to confirm, this is just for completeness (eg. the client won't break without this change)?

Yep. This allows any consumers of the client to have access to the `ProjectId` on the task as well if they want. Without this change they can still get and use tasks but it won't have `ProjectId`.

PR #850, opened 2024-05-17
Add SkipInitialHealthCheck flag to MachineBasedResource

scme0:
# Background

There are a few tests in Server which use Octopus.Server.Client to create machines but they don't want background processes like HealthChecks to run. Adding this property allows us to skip the initial health check when adding a machine to make testing easier, more stable and decrease the load on the server running in the test.

# Results

In the Client all we have to do is add the additional property. The server PR (https://github.com/OctopusDeploy/OctopusDeploy/pull/24785) will handle this flag and ensure an initial health check isn't executed if the flag is true.

### DIFF
--- a/source/Octopus.Server.Client/Model/MachineBasedResource.cs
+++ b/source/Octopus.Server.Client/Model/MachineBasedResource.cs
@@ -1,5 +1,4 @@
-﻿using System.Linq;
-using Octopus.Client.Extensibility;
+﻿using Octopus.Client.Extensibility;
 using Octopus.Client.Extensibility.Attributes;
 using Octopus.Client.Model.Endpoints;
 
@@ -51,5 +50,8 @@ namespace Octopus.Client.Model
         public string Architecture { get; set; }
         
         public string Slug { get; set; }
+
+        [Writeable]
+        public bool SkipInitialHealthCheck { get; set; }
     }
 }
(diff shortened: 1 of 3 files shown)


source/Octopus.Server.Client/Model/MachineBasedResource.cs
```diff
@@ -51,5 +50,8 @@ public abstract class MachineBasedResource : Resource, INamedResource, IHaveSlug
         public string Architecture { get; set; }
         
         public string Slug { get; set; }
+
+        [Writeable]
+        public bool SkipInitialHealthCheck { get; set; }
```
scme0, line 55: The only way I could get this information through the Client is by adding it to the MachineBasedResource. I don't love this but I also feel like the benefits outweigh the drawbacks in this case. Most people won't even care about this property.

I did consider adding a code comment above it to explain that it's only used when creating a new Machine (Deployment Target or Worker).

scme0: @APErebus I'm interested in your opinion on this.

APErebus: Do machines also have this initial health-check?

scme0: Yes, I believe they do, Although your comment made me realise I forgot to update the `CreateWorkerCommandHandler` in server.

scme0: I've updated the Server codebase to also pass this flag down when a worker is created.

It's interesting that the same test doesn't fail very often (it has failed the same way once) when it uses a worker instead of a deployment target. I believe this is because there is a second level of indirection with the event that causes a health check after creating a worker. 

When you create a worker a `MachineCreatedDomainEvent` is created which is handled in `MachineCreatedDomainEventMapper`. This mapper returns a `MachineCreatedEventV2` which is the same event directly raised when creating a deployment target.

APErebus (approved):

PR #860, opened 2024-06-26
Add SpaceId to MachineResource

egorpavlikhin:
For some reason `MachineResource` was missing `SpaceId` which could lead to an error where machines would be added to the default space instead of the one specified.

### DIFF
--- a/source/Octopus.Server.Client/Model/WorkerResource.cs
+++ b/source/Octopus.Server.Client/Model/WorkerResource.cs
@@ -35,7 +35,5 @@ namespace Octopus.Client.Model
             WorkerPoolIds.Clear();
             return this;
         }
-
-        public string SpaceId { get; set; }
     }
 }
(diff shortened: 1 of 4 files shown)


source/Octopus.Server.Client/Model/WorkerResource.cs
```diff
@@ -35,7 +35,5 @@ public WorkerResource ClearWorkerPools()
             WorkerPoolIds.Clear();
             return this;
         }
-
-        public string SpaceId { get; set; }
```
david-wu-octopus, line 39: What's the reason we're removing the `SpaceId` from `WorkerResource`?

david-wu-octopus (approved):
I left a comment. Otherwise, looks good to me!

source/Octopus.Server.Client/Model/WorkerResource.cs
egorpavlikhin: Because it's already inherited from MachineResource.

david-wu-octopus: Got it. Makes sense

PR #865, opened 2024-07-17
Add support for multiple Git dependencies

MissedTheMark:
We're adding support for non-primary Git dependencies: this PR adds client support for it.

The `GitResourceReferenceName` indicates whether or not it's a primary resource: no name makes it primary, specifying a name makes it non-primary.

See the equivalent property for Packages here: 
https://github.com/OctopusDeploy/OctopusClients/blob/f6d191220b086ebdc4c2684ed8e1296c790fffca/source/Octopus.Server.Client/Model/SelectedPackage.cs#L45

### DIFF
--- a/source/Octopus.Server.Client/Model/SelectedGitResource.cs
+++ b/source/Octopus.Server.Client/Model/SelectedGitResource.cs
@@ -25,4 +25,10 @@ public class SelectedGitResource
     /// The selected Git reference
     /// </summary>
     public GitReferenceResource GitReferenceResource { get; set; }
+    
+    /// <summary>
+    /// The name of the Git resource reference
+    /// </summary>
+    /// <remarks>May be empty for steps which have a primary Git resource</remarks>
+    public string GitResourceReferenceName { get; set; } = "";
 }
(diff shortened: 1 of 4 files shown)


source/Octopus.Server.Client/Model/SelectedGitResource.cs
```diff
@@ -25,4 +25,10 @@ public SelectedGitResource(string actionName, GitReferenceResource gitReferenceR
     /// The selected Git reference
     /// </summary>
     public GitReferenceResource GitReferenceResource { get; set; }
+    
+    /// <summary>
+    /// The name of the Git resource reference
+    /// </summary>
+    /// <remarks>May be empty for steps which have a primary Git resource</remarks>
+    public string GitResourceReferenceName { get; set; } = "";
```
APErebus, line 33: What does adding this allow us to do?

MissedTheMark: This makes the difference between primary and secondary (etc) git resources.   Providing this name makes it non-primary, leaving it blank makes it a primary.

See the equivalent property for Packages here: 
https://github.com/OctopusDeploy/OctopusClients/blob/f6d191220b086ebdc4c2684ed8e1296c790fffca/source/Octopus.Server.Client/Model/SelectedPackage.cs#L45

APErebus (approved):
Thanks for the clarification!

PR #872, opened 2024-09-17
Change HttpRouteExtractor access modifier to public to match interface

nathanwoctopusdeploy:
`OctopusAsyncClient` takes an `IHttpRouteExtractor` in its constructor, but the default implementation `HttpRouteExtractor` is not public.

Changing this class to public makes any inheritance of `OctopusAsyncClient` (and `HttpRouteExtractor` itself) much simpler.

Clones https://github.com/OctopusDeploy/OctopusClients/pull/871

[sc-91741]


500error (approved):
LGTM

PR #891, opened 2024-11-06
Add Git Repository Trigger

lourens-octopus:
[sc-85541]

Add Git Repository Trigger


geofflamrock (approved):
Looks good and lines up with the equivalent changes in Server. Looks like there might be some failing tests around public surface area that need looking into.

PR #897, opened 2024-11-25
Adds Generic OIDC accounts

IsaacCalligeros95:
This change adds support for generic Oidc accounts to the client for [#project-generic-oidc](https://octopusdeploy.slack.com/archives/C081K9EPXT9)

### DIFF
--- a/source/Octopus.Server.Client/Model/Accounts/GenericOidcAccountResource.cs
+++ b/source/Octopus.Server.Client/Model/Accounts/GenericOidcAccountResource.cs
@@ -13,11 +13,5 @@ namespace Octopus.Client.Model.Accounts
 
         [Writeable]
         public string[] DeploymentSubjectKeys { get; set; }
-
-        [Writeable]
-        public string[] HealthCheckSubjectKeys { get; set; }
-
-        [Writeable]
-        public string[] AccountTestSubjectKeys { get; set; }
     }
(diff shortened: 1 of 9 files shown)


source/Octopus.Server.Client/Model/Accounts/GenericOidcAccountResource.cs
```diff
@@ -0,0 +1,23 @@
...
+        public string Audience { get; set; }
+
+        [Writeable]
+        public string[] DeploymentSubjectKeys { get; set; }
+
+        [Writeable]
+        public string[] HealthCheckSubjectKeys { get; set; }
+
+        [Writeable]
+        public string[] AccountTestSubjectKeys { get; set; }
```
benPearce1, line 21: This is going to be confusing for users, I know we excluded it on the UI but this allows the user to set something that has no discernable effect

benPearce1 (approved):
This is good, but I think we should remove the health and account test subject keys

IsaacCalligeros95:
> This is good, but I think we should remove the health and account test subject keys

To prevent the resource mismatch errors I'd have up remove these fields from the resource in Server. I can do that but we'll end up with the mismatch being between the `GenericOidcAccountDetails` and Resource. Happy to remove this here if you'd prefer that.

PR #900, opened 2025-01-16
Add Config as Code Runbooks support

dylanlerch:
Adds support for CaC Runbooks to the client library. The new CaC Runbook methods are not returned in links collections, so these methods are a little different to the older ones, and you now need to pass the Project resource, Git ref and slug to all of these new calls.

For now, we're only adding these new methods to the asynchronous repositories. We may expand these to the others eventually if there is enough demand for it.

## Runbook and Runbook process repository changes
The bulk of the changes are in the `RunbookRepository` and `RunbookProcessRepository` as this is where all of the new methods requiring Git references will live. 

### Running a CaC Runbook
You no longer need to create a snapshot and run directly when running a CaC Runbook. Instead just call the Run method for a given Git ref and slug, and the snapshot and run will be created in the background automatically. 

Once a CaC Runbook is executed, the snapshots, runs, and tasks that are created are pretty much exactly the same, so no major changes should be required there, with the exception of retries.k

### Runbook retries
CaC Runbooks are always executed directly from a Git ref (rather than creating and running a snapshot), so we've also added a new method to the `RunbookRunRepository` to allow you to re-run these Runbooks. While these snapshots still exist in the API, but you can't create or run them directly. If you need to retry a failed Runbook for any reason, you call `Retry` on the run to execute that exact run again (rather than creating a new run from the snapshot). 

## Testing
For folks with internal access, there is a PR with tests against the latest version of Octopus here: https://github.com/OctopusDeploy/OctopusDeploy/pull/30513

[sc-95377]

### DIFF
--- a/source/Octopus.Server.Client/Model/ModifyRunbookProcessCommand.cs
+++ b/source/Octopus.Server.Client/Model/ModifyRunbookProcessCommand.cs
@@ -1,0 +1,7 @@
+namespace Octopus.Client.Model
+{
+    public class ModifyRunbookProcessCommand : RunbookProcessResource, ICommitCommand
+    {
+        public string ChangeDescription { get; set; }
+    }
+}
--- a/source/Octopus.Client.Tests/PublicSurfaceAreaFixture.ThePublicSurfaceAreaShouldNotRegress..NETCore.approved.txt
+++ b/source/Octopus.Client.Tests/PublicSurfaceAreaFixture.ThePublicSurfaceAreaShouldNotRegress..NETCore.approved.txt
@@ -3354,6 +3354,21 @@ Octopus.Client.Model
     Octopus.Client.Model.PersistenceSettingsType Type { get; }
     String Url { get; set; }
   }
+  class GitRunbookRunParameters
+  {
+    .ctor(String)
+    String Comments { get; set; }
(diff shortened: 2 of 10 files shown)


source/Octopus.Server.Client/Model/ModifyRunbookProcessCommand.cs
```diff
@@ -0,0 +1,7 @@
+namespace Octopus.Client.Model
+{
+    public class ModifyRunbookProcessCommand : RunbookProcessResource, ICommitCommand
```
geofflamrock, line 3: This type is only used for Git runbook processes, is that right? I wonder if its worth putting `Git` in the name e.g. `ModifyGitRunbookProcessCommand`. Same for the other one above.

geofflamrock (approved):
Looks good, just one question on a couple of type names

source/Octopus.Server.Client/Model/ModifyRunbookProcessCommand.cs
dylanlerch: The naming is consistent with deployment process, deployment settings, etc, which is why I went with this specific name. I see an argument either way.

PR #902, opened 2025-01-28
Pin FluentAssertions to v7

YuKitsune:
[sc-101162]

FluentAssertions v8 changes to a non-open-source license; Commercial Use requires a paid subscription at $129 USD per developer per year.

This PR pins the version of FluentAssertions to 7.0.0.

### DIFF
--- a/source/Octopus.Client.E2ETests/Octopus.Client.E2ETests.csproj
+++ b/source/Octopus.Client.E2ETests/Octopus.Client.E2ETests.csproj
@@ -25,7 +25,7 @@
         <PackageReference Include="NUnit" Version="4.2.2" />
         <PackageReference Include="NUnit3TestAdapter" Version="4.6.0" />
         <PackageReference Include="TeamCity.VSTest.TestAdapter" Version="1.0.41" />
-        <PackageReference Include="FluentAssertions" Version="6.12.2" />
+        <PackageReference Include="FluentAssertions" Version="[7.0.0]" />
     </ItemGroup>
 
     <ItemGroup Condition=" '$(TargetFrameworkIdentifier)' == '.NETFramework' ">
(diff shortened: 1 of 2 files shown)


source/Octopus.Client.E2ETests/Octopus.Client.E2ETests.csproj
```diff
@@ -25,7 +25,7 @@
         <PackageReference Include="NUnit" Version="4.2.2" />
         <PackageReference Include="NUnit3TestAdapter" Version="4.6.0" />
         <PackageReference Include="TeamCity.VSTest.TestAdapter" Version="1.0.41" />
-        <PackageReference Include="FluentAssertions" Version="6.12.2" />
+        <PackageReference Include="FluentAssertions" Version="[7.0.0]" />
```
borland, line 28: I reckon if we're going to lock a version like this we should pick 7.1.0 which is the current latest v7 release

borland: I did some similar tests (see [thread here](https://octopusdeploy.slack.com/archives/CDANN5QLT/p1738021287718839?thread_ts=1736889564.178519&cid=CDANN5QLT))

It's not clear that will help.

Rider will quite happily offer to trample on a pinned version or range and update FluentAssertions to the latest 8.0.1 (I also tried with `AllowedVersions="[,8.0.0)"`

So it comes down to whether Dependabot and/or Renovate respect the Nuget range syntax, either in the Version or AllowedVersions field. I don't know if they do, and I haven't put in the effort to do further tests.

Let's follow this up in [#incident-20250123-fluentassertions-is-no-longer-free](https://octopusdeploy.slack.com/archives/C08A5UARGUC)

borland (approved):
While the the `[7.1.0]` might not be the ideal solution, it's what the other teams are doing; good stuff, let's merge this and worry about finer detail later.

YuKitsune:
Closing this as Dependabot, Renovate and Rider do not respect NuGet version ranges in `csproj` files.
We should instead configure our dependency upgrader (Dependabot/Renovate) to prevent upgrades beyond v7.

PR #906, opened 2025-02-12
Retrieve default/missing tenant common variables

bec-callow-oct:
[SC-103509]

Required for: https://github.com/OctopusDeploy/OctopusDeploy/pull/31008

Previously, all templates were returned by the Get Tenant Variables endpoint which allowed Octopus UI and the Go CLI to find any missing/default values. When creating the new `api/{SpaceId}/tenants/{TenantId}/commonvariables` endpoint, we removed this functionality and only included templates for assigned variables. This PR adds a new array of missing and default variables to ensure that users can continue to access this data as required.

Default values are used for variables where a common variable template or project variable template has a default value set, but no value is explicitly set for one or more environment scopes.

Missing variables are similar to default variables, but are identified as having no default value for the template.

The new IncludeMissingCommonVariables boolean flag allows users to exclude missing variables if they are not required to avoid unnecessary queries by Octopus Server.

### DIFF
--- a/source/Octopus.Server.Client/Model/TenantVariables/ModifyCommonVariablesByTenantIdCommand.cs
+++ b/source/Octopus.Server.Client/Model/TenantVariables/ModifyCommonVariablesByTenantIdCommand.cs
@@ -15,8 +15,12 @@ public class TenantCommonVariable(string libraryVariableSetId, string templateId
     public string Id { get; set; } = string.Empty;
 
     public string LibraryVariableSetId { get; set; } = libraryVariableSetId;
+    
+    public string LibraryVariableSetName { get; set; }
 
     public string TemplateId { get; set; } = templateId;
+
+    public ActionTemplateParameterResource Template { get; set; }
 
     public PropertyValueResource Value { get; set; } = value;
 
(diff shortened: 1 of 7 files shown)


source/Octopus.Server.Client/Model/TenantVariables/ModifyCommonVariablesByTenantIdCommand.cs
```diff
@@ -15,9 +15,13 @@ public class TenantCommonVariable(string libraryVariableSetId, string templateId
     public string Id { get; set; } = string.Empty;
 
     public string LibraryVariableSetId { get; set; } = libraryVariableSetId;
+    
+    public string LibraryVariableSetName { get; set; }
```
tothegills, line 19: This class may need to be split as it currently applies to both the response and modify command. Unless it is required by the modify command?

tothegills (commented):
This is looking good, I like that you've made the new properties optional to keep the response slim by default.

Just the inline question/comment to consider.

source/Octopus.Server.Client/Model/TenantVariables/ModifyCommonVariablesByTenantIdCommand.cs
bec-callow-oct: Good point. It's no needed for the command so I'll make the change

tothegills (approved):

PR #908, opened 2025-02-17
Retrieve default/missing tenant project variables

bec-callow-oct:
[SC-103510]

Previously, all templates were returned by the Get Tenant Variables endpoint which allowed Octopus UI and the Go CLI to find any missing/default values. When creating the new api/{SpaceId}/tenants/{TenantId}/projectvariables endpoint, we removed this functionality and only included templates for assigned variables. This PR adds a new array of missing and default variables to ensure that users can continue to access this data as required.

Default values are used for variables where a project variable template or project variable template has a default value set, but no value is explicitly set for one or more environment scopes.

Missing variables are similar to default variables, but are identified as having no default value for the template.

The new IncludeMissingProjectVariables boolean flag allows users to exclude missing variables if they are not required to avoid unnecessary queries by Octopus Server.


tothegills (approved):
Consistent with the common project variable PR

PR #913, opened 2025-03-20
Added semaphore to control number of requests

droyad:
This adds a semaphore around the request sending for both the sync and async client. This is used to limit the number of requests that are made so that you don't DOS the Octopus Server.

The [alternate](https://github.com/OctopusDeploy/OctopusClients/pull/912) was rejected as it didn't apply to sync client and 
HTTP/2

I manually tested both clients and changing the value affects the duration of the requests.

I tested it out against our internal server and when limited to 30 connections it's almost as fast as when it's 100, showing limited benefit from increased parallelism.


borland (approved):
Simple and straightforward, looks good to me. Once you've fixed the PublicSurfaceArea failing test feel free to merge

PR #920, opened 2025-04-29
🧹 Added dependabot and implemented hash pinning for third party actions

enf0rc3:
Added dependabot to check github actions and pinned third party actions, have left the default github and the internal actions.

Dependabot will now run weekly for this repo and check the github actions for updates

Also updated dorny/test-reporter to v2, and pinned release.

Before Merge:

- [x] ❓ We publish extensive test reports, Have we confirmed that these are not used elsewhere, and we are happy with the v2 change that will no longer publish them in the same way, but add to the summary.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,8 @@
+version: 2
+updates:
+
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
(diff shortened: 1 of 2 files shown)


borland (approved):
LGTM. I have some background here -- we don't do anything with the test results other than look at them on GitHub if a test fails. There's no downstream systems which could break with the test-reporter move :+1:

PR #925, opened 2025-05-18
Add missing packages for release

mik-ky:
Updates client to support getting missing packages for a release.

[sc-110352] [sc-110783]

### DIFF
--- a/source/Octopus.Server.Client/Repositories/Async/ReleaseRepository.cs
+++ b/source/Octopus.Server.Client/Repositories/Async/ReleaseRepository.cs
@@ -60,8 +60,6 @@ namespace Octopus.Client.Repositories.Async
         Task<LifecycleProgressionResource> GetProgression(ReleaseResource release);
         Task<LifecycleProgressionResource> GetProgression(ReleaseResource release, CancellationToken cancellationToken);
 
-        [Obsolete("Please use the overload with cancellation token instead.", false)]
-        Task<GetMissingPackagesForReleaseResponse> GetMissingPackagesForRelease(GetMissingPackagesForReleaseRequest request);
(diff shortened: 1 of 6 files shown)


source/Octopus.Server.Client/Repositories/Async/ReleaseRepository.cs
```diff
@@ -59,6 +59,10 @@ public interface IReleaseRepository : IGet<ReleaseResource>, ICreate<ReleaseReso
         [Obsolete("Please use the overload with cancellation token instead.", false)]
         Task<LifecycleProgressionResource> GetProgression(ReleaseResource release);
         Task<LifecycleProgressionResource> GetProgression(ReleaseResource release, CancellationToken cancellationToken);
+
+        [Obsolete("Please use the overload with cancellation token instead.", false)]
+        Task<GetMissingPackagesForReleaseResponse> GetMissingPackagesForRelease(GetMissingPackagesForReleaseRequest request);
```
tothegills, line 64: Why is there an implementation without a cancellation token?

tothegills (commented):
Looks good, just one question inline. I see there are other obsolete methods without a token but keen to understand why.

source/Octopus.Server.Client/Repositories/Async/ReleaseRepository.cs
mik-ky: I think it was there for backwards compatibility for existing uses, but I thought I'd keep it in there to be consistent even though ours is a new endpoint. Should I remove it?

mik-ky: Actually on second thought I think it makes more sense to remove it. I'll do that

tothegills (approved):

PR #929, opened 2025-05-29
Convert PackageAcquisitionLocation enum to TinyType 

enf0rc3:
Renamed PackageAcquisitionLocationResource to match server.


borland (approved):
As discussed on zoom, this looks good.
I'm not 100% sure about the pattern where `PackageAcquisitionLocation` derives from `PackageAcquisitionLocationResource` so we can do a "graceful rename", but let's take a punt on it. Approved

PR #933, opened 2025-06-03
v15 breaking changes

APErebus:
The feature branch will roll up multiple breaking changes into a single major version bump


droyad (approved):
The two PRs that contribute to this have been approved, so :shipit:

PR #935, opened 2025-06-09
Remove the `Octopus.Server.MessageContracts.Base` reference 

droyad:
It is no longer used by anything in the client.

The `Octopus.Client` assembly uses IL Inlining so that it can be standalone (`Octopus.Server.Client` does not).

I used dotPeak to compare this branch built assembly against the current release version and found the expected  differences:

image


borland (approved):
The ILMerge/repack changes make sense given that we now only have one "Octopus" assembly. It's great to clean up the unneeded project reference. Approved!

PR #936, opened 2025-06-09
Move Extension configurations into this project

droyad:
Previously the configuration for the extensions were all out in 11 repositories generating a `Octopus.Client.Extensibility.*` nuget package each. 

Initially I copied them out, preserving the structure, but then the server test alterted me that there were mismatches between the server and client resources. I couldn't find any reason for this, so I changed these resources to match the server resources.

The server side of this is https://github.com/OctopusDeploy/OctopusDeploy/pull/33878. 

One this is merged in I will:
- Update the server PR with the release version of the client and merge it
- Merge the docs PR https://github.com/OctopusDeploy/docs/pull/2746
- Mark all the relevant NuGet packages as deprecated

## More Detail

The Authentication clients were copied into the OctopusServer codebase, but we did not setup the ability to publish these as NuGet packages and therefor have not published any since. Also since the "extensions" are now no longer loaded dynamically, every version of Octopus Server had these extensions. Therefor it makes sense for these to be part of the core client.

The IssueTracker clients seemed to have been archived and not copied anywhere. This brings them back into a maintained code base.

The ITSM client contacts were non-existant outside of the tests. Added those as well.


APErebus (approved):
LGTM!

PR #942, opened 2025-07-23
Fake Pull Request for Test Purposes

borland:
This PR was automatically synchronised from an OctopusDeploy internal git repository using Copybara.

## Adding a Release Note

Any change to the OctopusClients code must have a release note.
Only changes that are internal-only (for example adding some unit tests) should not have release notes.

To add a Release Note, **Add a comment to this pull request** with the text "Release Note: &lt;your text here&gt;"

## Semantic Versioning

You must check exactly one of these options:

- [ ] Breaking Change (the Major version digit should be incremented)
- [ ] New Feature (the Minor version digit should be incremented)
- [ ] Other (the Patch version digit will be incremented)

## How to Merge this PR

Add a comment with the text "/merge". An automated process will merge the PR. 
This will fail if there are no release notes, as release notes are generally required.

Add a comment with the text "/merge norelease" to merge without release notes for internal-only changes.
Merging this way will not release the change (No package will be published to NuGet)


azulo-dev:
Merge

PR #944, opened 2025-07-24
Create Github Release on build

enf0rc3:
# Background

We want to provide release notes of the changes made to octopus clients.

This will help customers better understand any breaking changes or change between versions to help identify any potential regressions.

# Results

Added a Github action that will create a 'Github Release' when the project is built and triggered via a workflow when "release notes" are included.

In this PR, there is nothing that triggers the creation of release notes, but it does allow for manual triggering with "release notes" as an option.

Releases will not be created when:
- Triggered by PR (e.g. any PR, as it will append a `-` + branchname to the version)
- Schedule triggers (e.g. nightly "master" builds as a earlier step adds a `-` to the version that we skip)
- Manual 'run workflow ' when there are no "release notes" provided.

We are adding this feature in a essentially "dormant state" to enable a future piece of work that will trigger this workflow and pass in the release notes generated at merge.

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -21,7 +21,7 @@ on:
   workflow_dispatch:
     inputs:
       releaseNotes:
-        description: 'Custom release notes to prepend to the release body, release notes are only created if this is not empty.'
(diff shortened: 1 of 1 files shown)


enf0rc3:
Spike no longer required, has bee implemented separately

PR #946, opened 2025-07-28
Create Github release on build using commit messages

enf0rc3:
# Background
We want to provide release notes of the changes made to octopus clients.

This will help customers better understand any breaking changes or change between versions to help identify any potential regressions.

# Results
Added a Github action step that will create a 'Github Release' when the project is built and the latest commit contains `release notes:` with some release notes (any string) following.

E.g. 
If a commit message contains "release note: This is a major update with new features", the text "This is a major update with new features" will be extracted and included in the GitHub release.

This supports multi line:

The following commit message: 
image

Results in:

image

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -19,7 +19,7 @@ on:
 
    # Allows you to run this workflow manually from the Actions tab
   workflow_dispatch:
- 
+
 # Pass branch and patch number to Nuke OctoVersion
 # (for pull_request events we override the /refs/pull/xx/merge branch to the PR's head branch)
 env:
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yaml
```diff
@@ -236,3 +247,72 @@ jobs:
...
+            const releaseNoteMarker = 'release notes:';
+            const releaseNoteIndex = commitMessage.toLowerCase().indexOf(releaseNoteMarker);
+            
+            if (releaseNoteIndex !== -1) {
+              releaseNotes = commitMessage.substring(releaseNoteIndex + releaseNoteMarker.length).trim();
+              console.log(`Extracted release notes: ${releaseNotes}`);
+            }
+            
+            // Only create release if release notes are present
+            if (!releaseNotes || releaseNotes.trim() === '') {
```
enf0rc3, line 281: As discussed, update to handle the first release notes, and only on the single line, update to use regex match and match it to the regex on the other github actions

enf0rc3:
Spike no longer required

